PR #4, opened 2026-03-14
feat(index): configurable content paths for plugin indexing

NRG-Nad:
## Summary
- Adds `AdditionalContentPaths` setting to `UMonolithSettings` so plugin mount points (e.g. `/MyPlugin`, `/ShooterCore`) are indexed alongside `/Game`
- Adds `GetIndexedContentPaths()` and `IsIndexedContentPath()` static helpers
- Updates all 6 indexers (Animation, DataTable, Dependency, GameplayTag, Level, Niagara) to use configurable paths instead of hardcoded `/Game`

## Motivation
Projects using Game Feature Plugins or custom plugin content directories get no index coverage for assets outside `/Game`. This is a common pattern in Lyra-based projects and modular game architectures.

## Configuration
Via Project Settings > Plugins > Monolith > Indexing, or in `Config/MonolithSettings.ini`:
```ini
+AdditionalContentPaths=/MyPluginMountPoint
```

## Test plan
- [ ] Verify indexing with default settings (only `/Game`) still works
- [ ] Add a plugin content path and verify its assets appear in index queries
- [ ] Verify DependencyIndexer correctly filters cross-plugin dependencies

### DIFF
--- a/Source/MonolithCore/Private/MonolithSettings.cpp
+++ b/Source/MonolithCore/Private/MonolithSettings.cpp
@@ -8,3 +8,43 @@ const UMonolithSettings* UMonolithSettings::Get()
 {
 	return GetDefault<UMonolithSettings>();
 }
+
+TArray<FName> UMonolithSettings::GetIndexedContentPaths()
+{
+	TArray<FName> Paths;
+	Paths.Add(FName(TEXT("/Game")));
+
+	if (const UMonolithSettings* Settings = Get())
+	{
+		for (const FString& Path : Settings->AdditionalContentPaths)
+		{
+			if (!Path.IsEmpty())
+			{
+				Paths.AddUnique(FName(*Path));
+			}
+		}
+	}
+
+	return Paths;
+}
+
+bool UMonolithSettings::IsIndexedContentPath(const FString& PackagePath)
+{
+	if (PackagePath.StartsWith(TEXT("/Game/")))
+	{
+		return true;
+	}
+
+	if (const UMonolithSettings* Settings = Get())
+	{
+		for (const FString& ContentPath : Settings->AdditionalContentPaths)
+		{
+			if (!ContentPath.IsEmpty() && PackagePath.StartsWith(ContentPath + TEXT("/")))
+			{
+				return true;
+			}
+		}
+	}
+
+	return false;
+}
--- a/Source/MonolithCore/Public/MonolithSettings.h
+++ b/Source/MonolithCore/Public/MonolithSettings.h
@@ -35,6 +35,10 @@ public:
 
 	// --- Indexing ---
 
+	/** Content paths to index in addition to /Game. Add plugin mount points like /MyPlugin. */
+	UPROPERTY(config, EditAnywhere, Category="Indexing")
+	TArray<FString> AdditionalContentPaths;
+
 	/** Override path for ProjectIndex.db (empty = default Saved/ location) */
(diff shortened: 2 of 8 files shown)


tumourlove (changes requested):
## Overall

Really solid foundation here — clean, consistent pattern across all 7 indexer files, and it fills a genuine gap for GFP and Lyra-style modular projects. A few things to sort out before merging though.

---

## What's Good

**Consistent pattern throughout** — `GetIndexedContentPaths()` and `IsIndexedContentPath()` are implemented the same way across all 7 indexers. That kind of discipline makes the codebase much easier to reason about.

**`/Game` always first + deduplication** — `GetIndexedContentPaths()` puts `/Game` at the front and uses `AddUnique` for `AdditionalContentPaths`. Clean.

**The false-positive guard** — `IsIndexedContentPath()` appending a trailing `/` before calling `StartsWith` is a nice touch. Without it, `/Game` would match `/GameFeatures/Whatever` and you'd get weird surprises. Good catch.

---

## What Needs Fixing

### Main issue: the metadata pass still has the hardcoded filter

The post-pass indexers got updated, but `FIndexingTask::Run()` in `MonolithIndexSubsystem.cpp` still has:

```cpp
Filter.PackagePaths.Add(FName(TEXT("/Game")));
```

This is the enumeration pass that populates the DB in the first place. Since it only scans `/Game`, assets from `AdditionalContentPaths` never get inserted — so every `DB.GetAssetId()` call in the post-pass indexers returns `-1` and silently skips them. The setting effectively does nothing end-to-end right now.

The fix is the same pattern you already applied to the indexers — pull `GetIndexedContentPaths()` (or equivalent) and add each path to `Filter.PackagePaths`.

### Heads up: conflict incoming

We just landed automatic marketplace plugin indexing on `main` that touches `DependencyIndexer.cpp`, `LevelIndexer.cpp`, `MonolithSettings.h`, and `MonolithIndexSubsystem.cpp`. The two features are actually complementary — yours handles manually specified paths for GFPs/Lyra, ours auto-detects marketplace plugins via `IPluginManager`. They should work great together.

After fixing the main filter, we'll need to reconcile the two path-injection approaches. The cleanest outcome is probably merging `AdditionalContentPaths` into the same paths list alongside the auto-detected marketplace ones, so there's one place that owns "what gets indexed."

### Minor: tooltip on the setting

A short comment or tooltip on `AdditionalContentPaths` explaining the expected format (`/PluginMountPoint`, e.g. `/MyGameFeature`) would save users a headache. Worth adding before this ships.

---

## Summary

Fix the `FIndexingTask::Run()` filter and we're most of the way there. The rest is coordination with the marketplace indexing work. Good stuff overall — the concept is solid and the implementation is clean where it's been applied.

tumourlove:
Incorporated into master via manual merge (commit ba47191). We merged your `AdditionalContentPaths` setting, `GetIndexedContentPaths()` / `IsIndexedContentPath()` helpers, and the 4 indexer upgrades (Animation, DataTable, GameplayTag, Niagara).

We kept our `SetIndexedPaths()` approach for DependencyIndexer/LevelIndexer (landed earlier this session) and fixed the `FIndexingTask::Run()` metadata pass to include `AdditionalContentPaths` too — that was the bug we flagged in the review.

Your feature now works alongside our automatic marketplace plugin detection — users get both zero-config marketplace indexing AND manual path config for Game Feature Plugins / Lyra-style projects. Thanks for the great contribution! 🤝

PR #6, opened 2026-03-14
feat(source): project C++ source indexing and incremental pipeline

NRG-Nad:
## Summary
Adds the ability to index a UE project's C++ source into the same EngineSource.db used for engine symbols, enabling cross-references between project and engine code.

### Changes
- **Scripts/index_project.py** - standalone script that discovers all Source/ directories under a project root (top-level modules + plugin modules) and indexes them incrementally
- **IndexingPipeline.load_existing_symbols()** - loads the full symbol table into memory before incremental indexing so cross-references resolve
- **IndexingPipeline._finalize()** - now supports incremental mode: only extracts references from newly-indexed files
- Progress reporting with rate and ETA during reference extraction
- **schema.py init_db()** - safe to call on existing DB (handles readonly gracefully)
- **MonolithSourceSubsystem** - always reopens DB after reindex (supports external tools updating the DB)

## Usage
```bash
cd Plugins/Monolith/Scripts
python index_project.py --project /path/to/MyProject
```

## Test plan
- [ ] Run index_project.py on a project with multiple Source/ directories
- [ ] Verify project symbols appear in source_query results
- [ ] Verify cross-references between project and engine code resolve
- [ ] Verify incremental mode only processes new files during finalization

### DIFF
--- a/Scripts/index_project.py
+++ b/Scripts/index_project.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python3
+"""Index a UE project's C++ source into an existing Monolith EngineSource.db.
+
+Discovers all Source/ directories under the project root (top-level modules
+and plugin modules) and indexes them into the same database used for engine
+source, enabling cross-references between project and engine symbols.
+
+Usage:
+    python index_project.py                          # auto-detect project root
+    python index_project.py --project /path/to/MyGame
+    python index_project.py --db /path/to/EngineSource.db
+"""
+
+import argparse
+import sqlite3
+import sys
+import time
+from pathlib import Path
+
+# Allow importing from sibling package
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+
+from source_indexer.db.schema import init_db
+from source_indexer.indexer.pipeline import IndexingPipeline
+
+GAME_FEATURES_DIR = "Plugins/GameFeatures"
+
+
+def discover_modules(project: Path) -> list[tuple[Path, str, str]]:
+    """Find all Source/ dirs under the project."""
+    modules: list[tuple[Path, str, str]] = []
+
+    # Top-level Source/ (e.g. LyraGame, LyraEditor, MyGame)
+    top_source = project / "Source"
+    if top_source.is_dir():
+        for child in sorted(top_source.iterdir()):
+            if child.is_dir():
+                modules.append((child, child.name, "Project"))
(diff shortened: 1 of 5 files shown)


tumourlove (changes requested):
Hey! Really solid concept here — having project C++ symbols searchable alongside engine symbols via `source_query` is genuinely useful. The `discover_modules()` approach (top-level Source/ + recursive plugin Source/ + GameFeature detection) is well thought out, and the `_new_file_ids` tracking for incremental reference extraction is a smart optimization.

A few things that need addressing before this is ready to merge:

## schema.py silent error swallowing

The `except sqlite3.OperationalError` block in `init_db()` catches "readonly" errors and silently passes. The comment says "locked by another process" but the actual check is for "readonly" — those are different failure modes. If the DB happens to be on a read-only filesystem, `init_db` silently succeeds but all subsequent writes fail with confusing errors down the line. At minimum, logging a warning here would make debugging way easier.

## Error level downgrade in MonolithSourceSubsystem.cpp

The original code logs `Error` on indexing failure and only reopens the DB on success. This PR changes it to `Warning` and reopens regardless. The "may be expected if DB was built externally" reasoning makes sense for `index_project.py`, but a genuine engine indexing failure should still be an error. Suggestion: distinguish between "external DB already exists" (Warning + reopen) and "indexing genuinely failed" (Error, maybe don't reopen stale DB).

## Progress message change

The finalize message changed from `"Finalizing (inheritance + references)..."` to just `"Finalizing..."` — this affects all callers of `index_engine`, not just project indexing. The original was more informative. Either revert this or make the new per-file progress reporting replace it intentionally.

## Minor

`default_project = script_dir.parents[2]` assumes the script lives at `Plugins/Monolith/Scripts/`. A quick comment about the expected layout would help if someone ever moves the script.

Overall the architecture is clean and the incremental pipeline is a nice addition. Just needs the error handling tightened up. 👍

tumourlove:
Merged in commit fb00779. We applied the 3 fixes from the review in follow-up commit 119feef:

- `schema.py`: warning log on readonly DB instead of silent swallow
- `MonolithSourceSubsystem.cpp`: Error level for genuine failures, Warning only when DB was built externally
- `pipeline.py`: restored the informative finalize progress message

Project source indexing is a great addition — thanks for the contribution! 🤝

PR #9, opened 2026-03-30
Allow plugin to be deeper into Plugins/

fp12:
Hardcoded ProjectPluginsDir()/"Monolith"/"Saved" assumed the plugin was installed directly under Plugins/. When installed at a different path, Saved/ (EngineSource.db, ProjectIndex.db, .monolith_running) were created in a phantom Plugins/Monolith/ directory instead.

Replaced all four hardcoded occurrences with IPluginManager::Get() .FindPlugin("Monolith")->GetBaseDir()/"Saved", with the old path kept as a fallback. Added "Projects" module dependency to MonolithSource and MonolithEditor Build.cs.

### DIFF
--- a/Source/MonolithCore/Private/MonolithCoreModule.cpp
+++ b/Source/MonolithCore/Private/MonolithCoreModule.cpp
@@ -6,6 +6,7 @@
 #include "MonolithCoreTools.h"
 #include "Misc/FileHelper.h"
 #include "GenericPlatform/GenericPlatformProcess.h"
+#include "Interfaces/IPluginManager.h"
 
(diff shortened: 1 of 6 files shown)


tumourlove:
Nice catch — this is a real problem for anyone not installing at the default path. The IPluginManager approach with fallback is exactly right.

Worth noting there's ~14 more hardcoded paths in MonolithMesh (preset dirs, town gen) and `MonolithUpdateSubsystem::GetPluginPath()` that could use the same treatment. Happy to take a follow-up PR for those if you're interested, or we'll get to it.

Merging, thanks for the contribution!

PR #17, opened 2026-04-17
fix: prevent indexer OOM and reentrant texture compiler crash (#16)

### LINKED ISSUE #16: This plugin always scan and then crash.
No matter what I set it up, it keeps scanning. Even though it's the latest version, it keeps making you update (this can be turned off).

alakangas:
## Summary

Fixes the "plugin always scans and then crashes" symptom reported in #16 on large projects.

Two root causes identified:

1. **Reentrant texture compiler crash.** Deep-index batches loaded assets while `FTextureCompilingManager::ProcessAsyncTasks` was still running, re-entering the compiler.
2. **No memory ceiling.** Deep and post-pass indexers loaded assets without GC between batches, eventually exhausting physical RAM.

## Changes

- **New `MonolithMemoryHelper`** (header + impl): working-set / available-memory queries, forced GC, Slate-safe `YieldToEditor`, critical-memory check, package-unload hint.
- **Reentrant-compile guard**: `FAssetCompilingManager::Get().FinishAllCompilation()` is called before each deep-index batch, and in `DataTableIndexer` / `GASIndexer` before they walk assets.
- **Batched indexers**: `AnimationIndexer`, `LevelIndexer`, `MeshCatalogIndexer`, `NiagaraIndexer`, and the deep-index loop in `MonolithIndexSubsystem` now:
  - process in configurable batches
  - force GC between batches
  - throttle + yield when the memory budget is exceeded
  - emergency-pause when available memory is critical (<2 GB)
  - respect the async task notification Cancel button
- **New settings** under `Indexing > Performance`:
  - `MemoryBudgetMB` (default 24576)
  - `DeepIndexBatchSize` (default 8)
  - `PostPassBatchSize` (default 4)
  - `GCFrequencyBatches` (default 2)
  - `YieldTimeSeconds` (default 0.1)
  - `bDeferFirstTimeIndex` (default false)
  - `bLogMemoryStats` (default true)

No new modules, no new third-party deps. Purely internal to `MonolithIndex` + a settings struct in `MonolithCore`.

## Test plan

- [x] Full first-time index on a large project (~20 GB content) no longer OOMs the editor.
- [x] No reentrant-compile crash during indexing.
- [x] Memory working set stays under configured budget (observed GC cycles in log).
- [x] Editor remains responsive while indexing (Slate yield between batches).
- [x] `Cancel` button on the async notification stops the run cleanly.
- [ ] Reviewer: sanity-check default `MemoryBudgetMB = 24576` is appropriate — tuned for 32+ GB workstations; smaller machines may need to lower it.

Closes #16.

### DIFF
--- a/Source/MonolithCore/Public/MonolithSettings.h
+++ b/Source/MonolithCore/Public/MonolithSettings.h
@@ -129,6 +129,43 @@ public:
 	UPROPERTY(config, EditAnywhere, Category="Indexing")
 	bool bIndexMarketplacePlugins = true;
 
+	// --- Indexing Performance ---
+
+	/** Memory budget for indexing in megabytes. Indexing will pause and run GC when this limit is exceeded. */
+	UPROPERTY(config, EditAnywhere, Category="Indexing|Performance", DisplayName="Memory Budget (MB)",
+		meta=(ClampMin="1024", ClampMax="65536", ToolTip="Maximum memory usage during indexing before forcing garbage collection. Default 24GB - UE editor typically uses 8-12GB baseline."))
+	int32 MemoryBudgetMB = 24576;
+
+	/** Number of assets to process per batch during deep indexing. Lower values reduce memory spikes but increase indexing time. */
+	UPROPERTY(config, EditAnywhere, Category="Indexing|Performance", DisplayName="Deep Index Batch Size",
+		meta=(ClampMin="1", ClampMax="64", ToolTip="Assets processed per batch. Lower = less memory, slower indexing."))
+	int32 DeepIndexBatchSize = 8;
+
+	/** Number of assets to process per batch for post-pass indexers (levels, meshes). These are memory-heavy so use smaller batches. */
+	UPROPERTY(config, EditAnywhere, Category="Indexing|Performance", DisplayName="Post-Pass Batch Size",
+		meta=(ClampMin="1", ClampMax="32", ToolTip="Batch size for level/mesh indexing. Lower for large assets."))
+	int32 PostPassBatchSize = 4;
+
+	/** Run garbage collection every N batches during indexing. Lower values keep memory lower but slow down indexing. */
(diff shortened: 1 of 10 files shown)


tumourlove:
Hey @alakangas — massive thanks for this. The reentrant-compiler guard and the batched GC approach are both spot on, and the settings struct is clean.

Going to bundle this with one small addition before merging — auto-detect installed RAM and pick a memory-budget tier based on that. The 24 GB default will drop lower-spec machines (like @MAYLYBY's report in #16) straight back into OOM, so I want to get tier logic + scaled batch sizes in the same release. Happy to push it as a follow-up commit on top of your branch with full credit — your work is doing the heavy lifting here.

Also — heads up — I've had a MonolithAudio module (81 actions: Sound Cue + MetaSound graph building + batch ops, Phases 0-2 complete and tested) sitting locally that I straight-up forgot to push. Going to bundle that into the same release so the repo actually reflects what exists.

Re: #18 — once this lands with MonolithAudio visible, keen to take another pass together. The MetaSound indexer (graph walk → DB) looks genuinely additive to what MonolithAudio already does (Builder-API mutation), so there should be clean coexistence, just want to check the surface area. I'll tag you in the release notes either way.

07

tumourlove:
Merged and shipped in v0.13.0: https://github.com/tumourlove/monolith/releases/tag/v0.13.0

Bundled with a RAM tier auto-detect commit on top (for the low-spec case @MAYLYBY hit) and the MonolithAudio module push (81 actions — had been sitting local, figured this was the right time to get it out).

You're credited in the release notes. Huge thanks — the compile-manager guard + batched GC pattern is exactly what that indexer needed.

On #18 — now that MonolithAudio is visible, keen to take another pass. The MetaSound indexer (graph walk → DB) looks genuinely additive to what MonolithAudio already does (Builder-API mutation), so should be clean coexistence. Give me a day or two to read it properly and I'll come back with notes.

07

PR #33, opened 2026-04-23
fix(core): detect and retry silent HTTP bind failures on editor boot

MaxenceEpitech:
## Summary

`FHttpServerModule::StartAllListeners()` logs a clear error when it can't bind the socket — but doesn't surface it to callers. Monolith's `Start()` therefore claimed success on boot even when the plugin wasn't actually listening, leaving MCP clients stuck against a dead socket (commonly when the port is still held by a previous editor session).

## What changes

- **Verify the bind actually took** via a localhost TCP probe (`FSocket`) after `StartAllListeners`.
- **Retry on failure** with exponential backoff, up to 5 attempts (~7.5s worst case).
- **Do NOT call `StopAllListeners()` between retries.** `FHttpServerModule` only evicts a cached failed listener inside `GetHttpRouter()` when `bHttpListenersEnabled == true` (see `Engine/Source/Runtime/Online/HTTPServer/Private/HttpServerModule.cpp`, the "listeners are still enabled" branch). `StopAllListeners` flips that flag to false, so a subsequent `GetHttpRouter()` returns the cached broken listener and every retry ends up reusing the same dead socket. The retry path therefore only clears *our own* router/route state and lets `GetHttpRouter` evict the failed listener itself.
- **`FMonolithHttpServer::Restart(Port)`** unbinds our routes without disabling the module-level listener state, then reruns `Start`.
- **`Monolith.Restart` console command** — rebind the HTTP server from the editor without restarting the whole editor process.
- Adds `Sockets` + `Networking` module deps for the probe.

## Repro / symptom (from my setup)

UE logs on boot:
\`\`\`
LogHttpListener: Error: HttpListener unable to bind to 127.0.0.1:9316
LogMonolith: Monolith MCP server started on port 9316  ← lie, nothing is listening
\`\`\`

With this PR, the same cold-start log now reads:
\`\`\`
LogHttpListener: Error: HttpListener unable to bind to 127.0.0.1:9316  (attempt 1)
LogMonolith: HTTP bind attempt 2/5 on port 9316 — waiting 1.0s
LogMonolith: Monolith MCP server listening on port 9316 (attempt 2)
\`\`\`

## Test plan

- [x] Cold-start with port free → bind succeeds on attempt 1 (verified in my editor session).
- [ ] Cold-start with port held by a just-killed instance → retry path kicks in and succeeds within a few attempts (code review against UE source confirms the eviction logic, but I only validated the happy path at runtime).
- [ ] `Monolith.Restart` from the editor console after a failed bind → rebinds successfully.

### DIFF
--- a/Source/MonolithCore/MonolithCore.Build.cs
+++ b/Source/MonolithCore/MonolithCore.Build.cs
@@ -21,7 +21,9 @@ public class MonolithCore : ModuleRules
 			"Projects",
 			"AssetRegistry",
 			"EditorSubsystem",
-			"UnrealEd"
+			"UnrealEd",
+			"Sockets",
+			"Networking"
 		});
 	}
 }
--- a/Source/MonolithCore/Private/MonolithCoreModule.cpp
+++ b/Source/MonolithCore/Private/MonolithCoreModule.cpp
@@ -6,9 +6,19 @@
 #include "MonolithCoreTools.h"
 #include "Misc/FileHelper.h"
 #include "GenericPlatform/GenericPlatformProcess.h"
+#include "HAL/IConsoleManager.h"
 #include "Interfaces/IPluginManager.h"
 
 #define LOCTEXT_NAMESPACE "FMonolithCoreModule"
+
+// Console command: `Monolith.Restart` — stops and restarts the MCP HTTP server
+// on its configured port. Useful when the initial bind failed silently (e.g.
+// port held by a zombie instance at UE boot).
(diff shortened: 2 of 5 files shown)


tumourlove:
Merged in v0.14.3 — thanks @MaxenceEpitech. The retry loop + port probe approach is solid, and the StopAllListeners insight about UE's listener eviction was a good catch. Monolith.Restart console command confirmed working.

tumourlove:
Shipped in v0.14.3. Implemented locally rather than merged directly — thanks again.

PR #37, opened 2026-04-24
Add set_anim_graph_node_property, native-component set_component_property, extended HTTP bind retry

MaxenceEpitech:
## Summary

Three independent fixes to unblock common MCP-driven workflows that currently require manual editor intervention:

- **`animation.set_anim_graph_node_property`** — new action that mutates a property on the source `UAnimGraphNode`'s inner `FAnimNode` struct. Required because writes via `blueprint.set_cdo_property` are silently wiped on every AnimBP compile (the CDO is regenerated from graph nodes). This action edits the authoritative source, so changing e.g. `ModifyBone.BoneToModify.BoneName` or `RotationMode` actually persists.

- **`blueprint.set_component_property` on native inherited components** — previously only resolved BP-added components via `SimpleConstructionScript`, so setting a property on a native component (e.g. `ACharacter::Mesh`, or any custom `USceneComponent` declared in a C++ parent) returned "Component not found". Falls back to CDO default subobjects, and for `FObjectProperty` values (TObjectPtr fields like `SkinnedAsset` on `SkeletalMeshComponent`) replaces silent-failing `ImportText` with explicit `StaticLoadObject` + class compatibility check + `PostEditChangeProperty`, so setter-driven side effects run and the value is actually serialized to the .uasset.

- **HTTP bind retry extended** — on a close+reopen cycle, the OS keeps port 9316 in TIME_WAIT for up to ~30s and UE's `HttpServerModule` caches a failed listener internally. The old 5×~0.5s retry budget timed out every time, leaving the MCP server dead for the entire session. Bumps to 20×2s = 40s budget and calls `StopAllListeners()` between attempts to evict the stale listener cache.

All three issues were hit back-to-back while trying to script an FPS/TPS mesh split refactor entirely via MCP. With these three fixes, the refactor went from "blocked by editor-only steps" to fully scriptable end-to-end.

## Test plan

- [ ] `animation.set_anim_graph_node_property` on a `ModifyBone` node's `RotationMode` → persists across `compile_blueprint` + `save_asset` + re-read via `get_cdo_properties` on the subobject path
- [ ] `blueprint.set_component_property` on a native inherited `USkeletalMeshComponent`'s `SkinnedAsset` → re-read shows the loaded asset, PIE renders correctly
- [ ] `blueprint.set_component_property` on a native component with an abstract-base property class (e.g. `TObjectPtr<USkinnedAsset>`) → fallback load without class constraint + compatibility check works
- [ ] Close UE, immediately reopen → Monolith HTTP server binds within ~40s despite TIME_WAIT (log: `Monolith MCP server listening on port 9316 (attempt N)` where N < 20)

### DIFF
--- a/Source/MonolithAnimation/Private/MonolithAbpWriteActions.cpp
+++ b/Source/MonolithAnimation/Private/MonolithAbpWriteActions.cpp
@@ -102,6 +102,23 @@ void FMonolithAbpWriteActions::RegisterActions(FMonolithToolRegistry& Registry)
 			.Optional(TEXT("position_x"), TEXT("number"), TEXT("Node X position (default: 0)"), TEXT("0"))
 			.Optional(TEXT("position_y"), TEXT("number"), TEXT("Node Y position (default: 0)"), TEXT("0"))
 			.Build());
+
+	// --- set_anim_graph_node_property ---
+	// Mutates a property on the *source* UAnimGraphNode's inner FAnimNode struct.
+	// Writing via blueprint.set_cdo_property is wiped on compile because the AnimBP's
+	// CDO is regenerated from the graph nodes — this action edits the authoritative
+	// source so the change persists.
+	Registry.RegisterAction(TEXT("animation"), TEXT("set_anim_graph_node_property"),
+		TEXT("Mutate a property on an existing anim graph node's internal FAnimNode struct (e.g. ModifyBone.BoneToModify.BoneName, ModifyBone.RotationMode, TwoBoneIK.EffectorLocationSpace). Persists across compile — writes to the source UAnimGraphNode, not the CDO."),
+		FMonolithActionHandler::CreateStatic(&HandleSetAnimGraphNodeProperty),
+		FParamSchemaBuilder()
+			.Required(TEXT("asset_path"), TEXT("string"), TEXT("Animation Blueprint asset path"))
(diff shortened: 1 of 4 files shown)


tumourlove (approved):
Reviewed end-to-end. All three fixes target real, demonstrable defects:

1. `set_anim_graph_node_property` correctly writes to the source `UAnimGraphNode` instead of the CDO (which is regenerated on every compile). Heuristic struct discovery (any `FStructProperty` derived from `FAnimNode_Base`) survives subclass renames.

2. Native-component fallback in `set_component_property` is justified — native inherited components live as CDO default subobjects, not on the SCS. The `FObjectProperty` special-case (explicit `StaticLoadObject` + class compatibility check + `PostEditChangeProperty`) is the right call: ImportText is genuinely fragile for `TObjectPtr` on CDO subobjects with abstract-base property classes (e.g. `USkinnedAsset`), and `PostEditChangeProperty` fires the setter side effects (`SkinnedAsset` ↔ `SkeletalMesh` mirror, render-state refresh) that prevented values from being serialized to the .uasset.

3. HTTP bind retry extension is correct: the key fix is `FHttpServerModule::Get().StopAllListeners()` between attempts — the module caches the failed listener and won't rebind without it. 40s budget covers TIME_WAIT comfortably.

Plan is to smoke-test the new write paths post-merge against disposable assets in the same editor-up window as some unrelated CORS verification work.

Thanks for the contribution.

Issue #29, opened 2026-04-21
Data loss: save_asset silently drops cross-package TObjectPtr refs on Blueprint CDOs and nested struct arrays


danielandric:
# Data loss: `save_asset` silently drops cross-package `TObjectPtr` refs on Blueprint CDOs and nested struct arrays

  **UE 5.7 · Monolith v0.14.0 (also on 0.13.x)**

  Flagging this because it's silent data loss and I've pushed the investigation as far as I can without a UE source build. Would really appreciate someone with more engine-internals knowledge taking a look or suggesting angles I haven't tried.

  There are two symptoms that may or may not share a root cause. I investigated the first one deeply and localized it to `FLinkerSave`. The second one I observed and characterized but didn't get as far with.

  Baseline for both: normal editor workflow — Details panel edit + File→Save — preserves these refs correctly. The drop is specific to saves triggered through Monolith.

  ---

  ## Symptom 1 — Blueprint CDO cross-package `TObjectPtr` (investigated)

  `set_cdo_property` a `TObjectPtr<UX>` on a Blueprint CDO, where `UX` is an asset class and the target asset lives in another package (e.g. `/Game/Other/SomeAsset.SomeAsset`). Then `save_asset`. The write
  lands in memory — readback returns the resolved path, PIE works — but `save_asset` never writes an import-table entry for the referenced asset. Close the editor, reopen, the ref is `None` and a byte-scan of
  the `.uasset` confirms the path is gone.

  When Monolith's save runs on the Blueprint, every cross-package `TObjectPtr` asset ref on the CDO drops from the saved bytes — both the one I just wrote and any that were already there from prior
  Details-panel edits (which had survived previous editor saves fine). `TSubclassOf` class refs on the same CDO survive. So the broken path is object-ref encoding specifically, not class-ref encoding.

  I initially read the disappearance of the prior editor-set refs as `compile_blueprint` wiping overrides; it isn't. Compile preserves the values in memory. The drop happens at the save step, and
  `compile_blueprint` is not required to reproduce it (see rule-outs).

  ## Symptom 2 — nested `TArray<FStruct>` with a cross-package `TObjectPtr` inside (observed, shallower dig)

  Canonical case: `UInputMappingContext::Mappings` (`TArray<FEnhancedActionKeyMapping>`, each entry has `Action: TObjectPtr<UInputAction>`). `set_cdo_property` the whole array, save, reopen — every `Action` is
[...]

PR #43, opened 2026-04-26
fix(index): drop RF_Transient on UPackage in TryUnloadPackage (Issue #42)

### LINKED ISSUE #42: Indexer poisons packages with RF_Transient — strips cross-package refs at save (#29 carrier)
## Relationship to #29 / PR #39

This is the carrier mechanism behind the silent data loss reported in #29 ("Data Loss in `save_asset`"). PR #39 (v0.14.5) shipped four save-pipeline fixes — transient-outer inline subobjects in `FJsonObjectConverter`, cradle/reparent walker unification, `FMapProperty::ValueProp` double-offset, sparse-iteration with `GetMaxIndex()` — and **explicitly deferred poisoned-asset tooling** pending alignment with the actual `UObject::IsAsset()` spec.

This PR fixes the carrier that produces the deferred poisoning case: `FMonolithMemoryHelper::TryUnloadPackage` setting `RF_Transient` on the indexed asset's `UPackage`.

The #29 report noted that "`git checkout` of pre-damage bytes still produces broken saves on subsequent File→Save operations, even without Monolith involvement." That's the carrier in action: once the indexer has marked the in-memory `UPackage` with `RF_Transient` at startup, the flag persists in the live UObject regardless of file-system bytes. A clean `git checkout` doesn't load fresh bytes into the existing package object; it just changes what's on disk. The next save reads the still-poisoned in-memory state. Even cold-restart only buys a brief window — the indexer's first-time-index pass re-applies `RF_Transient` within seconds, so the bug returns before most users ever interact with the asset.

## TL;DR

`FMonolithMemoryHelper::TryUnloadPackage` (`Source/MonolithIndex/Private/MonolithMemoryHelper.cpp:84`) calls `Package->SetFlags(RF_Transient)` after a deep-index pass, intending to encourage GC. `RF_Transient` does not control GC — it's a save flag. When GC fails to reclaim the package (because the asset is still referenced by a BP CDO, editor watcher, or thumbnail cache), the live `UPackage` retains `RF_Transient`. `UObject::IsAsset()` then returns `false` for any asset in that package (UE 5.7 `Engine/Source/Runtime/CoreUObject/Private/UObject/Obj.cpp:2733`), and **cross-package `TObjectPtr` references to those assets are silently stripped at save time**. The corruption survives cold restart because the indexer re-applies the flag on every editor startup.

## Symptoms users observe

- Cross-package `TObjectPtr` property values get silently nulled when the containing asset is saved (Blueprint compile + save, Input Mapping Context default-mappings arrays, post-process material refs in level actors, etc.).
- `UEditorAssetSubsystem::SaveLoadedAsset` fails with `"Asset is not registered. The object 'X' is not an asset."` for affected targets.
- The corruption is class-conditional: assets whose `UClass` has a registered deep indexer are affected (`UInputAction` via `FInputActionIndexer`; `UMaterial` via `FMaterialIndexer`; `UStaticMesh` via `FMeshCatalogIndexer`; `UNiagaraSystem`; `UWorld`). `UInputMappingContext`, `UBlueprint`, `UDataAsset` subclasses are not directly poisoned (no indexer registration), but they can lose refs *to* poisoned targets at save.
- `UPackageTools::ReloadPackages` appears to fix the asset in-session, but cold-restart re-applies the poisoning when the indexer runs again.

## Affected configurations

Default settings poison. The issue triggers whenever any of these indexer flags is enabled (all default `true`):

- `bIndexInputActions`
- `bIndexMaterials`
- `bIndexMeshCatalog`
- `bIndexNiagara`
- `bIndexLevels`

## Reproduction

1. Cold-start the editor with default Monolith settings on a project containing at least one `UInputAction` and one Blueprint that references it via `TObjectPtr<UInputAction>`.
2. Wait for the first-time indexing pass to complete.
3. Open the Blueprint, modify the `TObjectPtr<UInputAction>` property (e.g. clear and re-pick the same IA, or pick a different one), compile, save. The save reports success.
4. Reopen the Blueprint or restart the editor. The `TObjectPtr<UInputAction>` value is `None`. No warning, no error was logged at save time.

## Diagnosis

After the first-time indexing pass completes, run any of these from a Blueprint Utility, Editor Utility, or `c++` console to confirm an asset has been poisoned:

```cpp
// Returns false on a poisoned asset; true on a healthy one.
const bool bIsAsset = Asset->IsAsset();

// Direct flag check on the package — the carrier symptom.
const bool bPackagePoisoned = Asset->GetOutermost()->HasAnyFlags(RF_Transient);
```

A poisoned asset reports `IsAsset() == false` and `GetOutermost()->HasAnyFlags(RF_Transient) == true`. Healthy assets in the same project (whose `UClass` has no registered deep indexer, or whose package was successfully GC'd between indexer pass and your check) report the inverse.

## Root cause

`MonolithMemoryHelper.cpp` lines 82–87 (pre-fix):

```cpp
// Mark the package as pending kill so it gets cleaned up on next GC
Package->ClearFlags(RF_Standalone);
Package->SetFlags(RF_Transient);   // ← carrier line

// Clear any references we might have
Asset->ClearFlags(RF_Standalone);
```

The `SetFlags(RF_Transient)` line was added in PR #17 (commit `168c087`, the OOM fix). The intent stated in the comment was to encourage GC reclamation. The mechanism is wrong. UE 5.7 source:

- `Engine/Source/Runtime/CoreUObject/Public/UObject/ObjectMacros.h:565` — `RF_Transient = "Don't save object."` Pure serialization flag.
- `Engine/Source/Runtime/CoreUObject/Public/UObject/GarbageCollection.h:28` — `#define GARBAGE_COLLECTION_KEEPFLAGS (GIsEditor ? RF_Standalone : RF_NoFlags)`. **GC's editor keep-list is exactly `RF_Standalone`.** `RF_Transient` is not consulted by GC. Searched `IncrementalGC | GarbageCollection.cpp | UObjectArray | GarbageCollectionInternal` for any `RF_Transient` reference — zero hits.
- `Engine/Source/Runtime/CoreUObject/Public/UObject/ObjectMacros.h:618` — `RF_PropagateToSubObjects` includes `RF_Transient`, so newly-constructed sub-objects under a poisoned package inherit the flag.

When the package is reclaimable (no other strong refs), GC removes it cleanly and the flag becomes irrelevant. When GC fails to reclaim — common, because packages can be pinned by Blueprint CDOs, the asset registry, the thumbnail cache, watchers, or any code holding a `TObjectPtr` — the live UPackage keeps `RF_Transient`. From that point until cold restart (and again after, because the indexer pass re-applies it), `UObject::IsAsset()` returns `false` per `Obj.cpp:2733`:

```cpp
return ObjectPackage != GetTransientPackage()
    && !ObjectPackage->HasAnyFlags(RF_Transient)
    && !ObjectPackage->HasAnyPackageFlags(PKG_PlayInEditor);
```

The save pipeline harvests cross-package references via paths that consult `IsAsset()`. References whose target reports `false` get silently stripped — no warning logged, no save error, no dirty-mark on the loss.

## Severity

Silent data corruption on the default configuration. The user's Blueprint compiles, saves cleanly, and the on-disk asset has nulled refs. Common workflows that hit this:

- Authoring a controller Blueprint that references Input Actions.
- Wiring post-process volumes / level actors that hold `TObjectPtr<UMaterial>`.
- Input Mapping Context default mappings (`FEnhancedActionKeyMapping.Action`).

## Fix

One line removed, comment expanded to call out the failure mode for the next developer who reaches for `RF_Transient` on a package. The GC-eligibility intent of the original code is fully delivered by `Package->ClearFlags(RF_Standalone)`, which is preserved.

PR #39's deferred "poisoned-asset tooling" scope is obsoleted by this fix: with the carrier removed, `IsAsset()` returns true for indexed assets and the deferred scan/repair tools are no longer needed for this failure mode.

danielandric:
Fixes #42.

## Summary

`FMonolithMemoryHelper::TryUnloadPackage` was setting `RF_Transient` on indexed-asset packages to "encourage GC". `RF_Transient` is a save flag, not a GC flag — when the post-batch `ForceGarbageCollection` fails to reclaim a package (any package still pinned by a BP CDO, watcher, asset registry, or thumbnail cache), the live `UPackage` retains `RF_Transient`, `UObject::IsAsset()` returns `false`, and cross-package `TObjectPtr` saves silently strip refs to those targets. This is the carrier behind #29's data loss and the deferred poisoned-asset case from PR #39.

The fix is one line: drop `SetFlags(RF_Transient)`. The GC-eligibility intent is preserved by `Package->ClearFlags(RF_Standalone)` — that's what `GARBAGE_COLLECTION_KEEPFLAGS` actually keys on in editor.

## Caller audit

`TryUnloadPackage` has 5 in-tree callers — `MonolithIndexSubsystem.cpp:672`, `MeshCatalogIndexer.cpp:134/233`, `LevelIndexer.cpp:97/119`, `NiagaraIndexer.cpp:77`. None reads the flag back, none branches on it, none saves the package later expecting `RF_Transient`. All follow the "mark → batch → `ForceGarbageCollection`" pattern. In-tree `RF_Transient` readers (`MonolithAssetUtils.cpp:102`, `MonolithMeshQualityActions.cpp:713`, ephemeral `NewObject<>(GetTransientPackage(), …, RF_Transient)` constructions) are all unrelated to indexed-asset packages.

## Residual risk surfaces (not fully closed)

- **Large-project OOM stress untested.** I have not stress-tested at the ~20 GB scale that motivated #17. The size-invariance argument is strong (`RF_Transient` is not on GC's keep-list at any scale), but it's an argument, not an experiment. If users with large content sets hit a regression after this lands, please reopen with telemetry.
- **Indirect / runtime-computed flag tests.** The audit was lexical on `RF_Transient` and `HasAnyFlags(...Transient)`. UE has flag-querying paths where the flag arg is computed at runtime; those wouldn't have been caught.
- **`RF_PropagateToSubObjects` includes `RF_Transient`.** Pre-fix, any `NewObject<T>(Package, …)` between an indexer pass and the next GC would have constructed `T` with `RF_Transient` inherited. I have not been able to construct a real Monolith path where this matters, but if downstream tooling expected those sub-objects to be transient, behavior changes.
- **Save-on-close / autosave.** Post-fix, indexer-touched packages that were dirtied (by anything) will save normally instead of being implicitly suppressed. This is desired behavior, not a regression — but it's a behavior change worth naming.

## Alternatives considered

- **Replace with `MarkAsGarbage()` or a stronger GC nudge.** Rejected: scope creep. `Package->ClearFlags(RF_Standalone)` is the canonical editor pattern for "let GC reclaim this" — it's literally what `GARBAGE_COLLECTION_KEEPFLAGS` keys on. Adding new behavior beyond restoring correctness invites an unrelated review fight.
- **Per-asset `repair_target` action instead of fixing the carrier.** Rejected: band-aid. Forces every user to run a repair tool, and `is_asset` flips back at the next cold restart.
- **Comment the line out instead of deleting it.** Rejected: git history preserves the line; commented-out code rots. The new comment explains what NOT to do, citing `Obj.cpp:2733`, so the next developer hitting "I want to encourage GC for this package" doesn't re-introduce it.

### DIFF
--- a/Source/MonolithIndex/Private/MonolithMemoryHelper.cpp
+++ b/Source/MonolithIndex/Private/MonolithMemoryHelper.cpp
@@ -79,11 +79,12 @@ bool FMonolithMemoryHelper::TryUnloadPackage(UObject* Asset)
 		return false;
 	}
 
-	// Mark the package as pending kill so it gets cleaned up on next GC
+	// Clear RF_Standalone on the package and asset so they're GC-eligible.
+	// Do NOT SetFlags(RF_Transient) on the package — RF_Transient doesn't
+	// control GC, and it causes UObject::IsAsset() to return false on contained
+	// assets (Obj.cpp:2733), which silently strips cross-package TObjectPtr refs
+	// at save time ("target poisoning").
 	Package->ClearFlags(RF_Standalone);
-	Package->SetFlags(RF_Transient);
-
-	// Clear any references we might have
 	Asset->ClearFlags(RF_Standalone);
 
 	return true;


tumourlove:
Good catch on the RF_Transient issue. The fix is clean - removing the flag instead of trying to work around it is the right call. Will be in the next release (this weekend).

tumourlove:

Thanks for the `RF_Transient`-on-UPackage root-cause fix, @danielandric! The "obvious-looking save flag, devastating GC consequences" diagnosis closed Issue #42 and a class of silent cross-package `TObjectPtr` corruption.

PR #44, opened 2026-04-26
feat(blueprint): add component-bound and multicast delegate nodes to add_node

danielandric:
## Summary

Adds five new `node_type` values to `blueprint.add_node` and `blueprint.resolve_node` covering UE's full delegate-binding right-click menu — `Bind Event to ...`, `Unbind Event from ...`, `Unbind all Events from ...`, `Call ...`, plus the green `+OnClicked`-style component-bound event entry.

- **`ComponentBoundEvent`** — the event-entry node spawned by clicking `+` next to a component delegate in Designer (e.g. `Button.OnClicked`). Validates that the component variable resolves on `BP->GeneratedClass` and the delegate is `BlueprintAssignable`. Rejects duplicate (component, delegate) pairs BP-wide via `FKismetEditorUtilities::FindBoundEventForComponent` (matches the editor's own dedupe semantics across ubergraph pages, not just per-graph). `FindComponentProperty` accepts UMG widget properties, not just `UActorComponent` subclasses, so widget Blueprints work alongside Actor Blueprints.
- **`AddDelegate`** — the "Bind Event to ..." node that runtime-binds an event to a `BlueprintAssignable` multicast delegate. Defaults to self-context (`BP->GeneratedClass`); `target_class` accepts both bare (`MyController`) and prefixed (`AMyController`) forms. `SetFromProperty` is called with `DelegateProp->GetOwnerClass()` so inherited delegates resolve to the declaring class, not the user-supplied target.
- **`RemoveDelegate`** — the "Unbind Event from ..." node. Same shape and parameters as AddDelegate.
- **`ClearDelegate`** — the "Unbind all Events from ..." node. Removes every bound listener at runtime in one call. Same parameters as AddDelegate; spawned node has no `Delegate` input pin (no event reference needed).
- **`CallDelegate`** — the "Call ..." node. Broadcasts a BP-resident multicast delegate to all bound listeners — used when the dispatcher itself is authored in Blueprint (not C++). Spawned node has one input pin per delegate signature parameter.

The four multicast-delegate types share validation logic via a single helper, `MonolithBlueprintInternal::ResolveDelegateOwnerAndProperty`. `target_class` prefix-normalization (bare / `A`-prefixed / `U`-prefixed / strip-and-retry) and `BlueprintAssignable` multicast property lookup live there. AddDelegate's add_node and resolve_node branches were refactored to use the helper alongside the new types; the four resolve_node branches collapse into a single `OR` matcher with a switch on `NodeType`.

`SerializeNode` extended with a `K2Node_ComponentBoundEvent` branch (more-specific cast wins over `K2Node_Event`) and a `K2Node_BaseMCDelegate` branch — the latter covers all four multicast delegate types uniformly via shared `DelegateReference`. `add_nodes_bulk` and `batch_execute` pick up every new type transparently — no dispatch-layer changes.

Mechanism + per-decision rationale lives in the two commit messages — this body is review context only.

## Test plan

Validated end-to-end against a downstream consumer project. ComponentBoundEvent and AddDelegate were authored against UMG buttons and a player-controller multicast delegate respectively; compile-clean and runtime-verified in PIE. The three new types (Remove/Clear/Call) were validated via `blueprint_query.resolve_node` dry-runs against an actor class with a `BlueprintAssignable` multicast delegate: each produces the correct UE node title, pin shape (Add/Remove have 4 pins including `Delegate` input; Clear/Call have 3 pins, no event-ref needed), and class. Error paths verified end-to-end — missing `delegate_property_name`, unknown delegate name, unknown class, and missing `target_class` + `asset_path` all produce correctly-labeled errors via the helper's `NodeTypeLabel` substitution. Prefix normalization tested with both bare and `A`-prefixed `target_class` forms.

## Non-goals (future PRs)

- **`K2Node_AssignDelegate`** — auto-spawns a paired `CustomEvent` in `PostPlacedNewNode`. The `add_node` factory uses an abbreviated `NewObject` → `AllocateDefaultPins` pattern that bypasses `PostPlacedNewNode` for every node type. Adding AssignDelegate requires extracting a canonical-spawn helper and migrating every existing factory branch — small refactor, but with a per-node-type validation surface (variable nodes snapshot category, K2Node_Event does delegate housekeeping, etc.). Held for a separate PR.
- **Convenience `event_name` on `AddDelegate`** that auto-spawns + wires a `CustomEvent` in one call. Caller does it in three `add_node` calls today; YAGNI given measured authoring volume.

### DIFF
--- a/Docs/specs/SPEC_MonolithBlueprint.md
+++ b/Docs/specs/SPEC_MonolithBlueprint.md
@@ -74,7 +74,7 @@
 **Node & Pin Operations (6)**
 | Action | Params | Description |
 |--------|--------|-------------|
-| `add_node` | `asset_path`, `graph_name`, `node_class`, `position` | Add a node to a graph. Accepts common aliases (e.g. `CallFunction`, `VariableGet`, `ComponentBoundEvent`, `AddDelegate`) and tries `K2_` prefix fallback for function calls |
+| `add_node` | `asset_path`, `graph_name`, `node_class`, `position` | Add a node to a graph. Accepts common aliases (e.g. `CallFunction`, `VariableGet`, `ComponentBoundEvent`, `AddDelegate`, `RemoveDelegate`, `ClearDelegate`, `CallDelegate`) and tries `K2_` prefix fallback for function calls |
 | `remove_node` | `asset_path`, `graph_name`, `node_id` | Remove a node by ID |
 | `connect_pins` | `asset_path`, `graph_name`, `source_node`, `source_pin`, `target_node`, `target_pin` | Connect two pins |
 | `disconnect_pins` | `asset_path`, `graph_name`, `source_node`, `source_pin`, `target_node`, `target_pin` | Disconnect two pins |
--- a/Docs/specs/SPEC_MonolithBlueprint.md
+++ b/Docs/specs/SPEC_MonolithBlueprint.md
@@ -74,7 +74,7 @@
 **Node & Pin Operations (6)**
 | Action | Params | Description |
 |--------|--------|-------------|
-| `add_node` | `asset_path`, `graph_name`, `node_class`, `position` | Add a node to a graph. Accepts common aliases (e.g. `CallFunction`, `VariableGet`) and tries `K2_` prefix fallback for function calls |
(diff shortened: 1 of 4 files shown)


tumourlove:
Tested ComponentBoundEvent on a widget button - works as expected. The delegate resolution and prefix normalization look good. Will be in the next release (this weekend).

tumourlove:

Thanks for the full delegate-node family, @danielandric! `ComponentBoundEvent`, `AddDelegate`, `RemoveDelegate`, `ClearDelegate`, `CallDelegate` — closes the asymmetry where Monolith only authored the editors Bind verb.

PR #46, opened 2026-04-27
fix(blueprint): bind add_event_node to inherited overrides on non-AActor classes

### LINKED ISSUE #47: add_event_node falls through to K2Node_CustomEvent for UUserWidget Tick (and other AActor-aliased names on non-AActor parents)
## TL;DR

`HandleAddEventNode` (`Source/MonolithBlueprint/Private/MonolithBlueprintNodeActions.cpp`) aliases AActor-style event names to their `ReceiveX` counterparts (`tick → ReceiveTick`, `beginplay → ReceiveBeginPlay`, etc.) before walking the parent class chain. Non-AActor `BlueprintImplementableEvent` hosts use the bare names — `UUserWidget` declares `Tick`, not `ReceiveTick`. The alias-resolved walk returns no match and the action silently falls through to the `K2Node_CustomEvent` branch, producing a custom event titled "Tick" that compiles but never fires on widget tick. Authoring widget Tick chains via `add_event_node` is blocked.

## Symptoms users observe

```python
add_event_node({asset_path: "/Game/UI/MyWidget", event_name: "Tick"})
```

returns:

```json
{
  "id": "K2Node_Event_3",
  "class": "CustomEvent",
  "title": "Tick",
  "is_override": false,
  ...
}
```

instead of the expected `K2Node_Event` with `is_override: true`, `class: "UserWidget"`, and the standard UMG Tick pin shape (`OutputDelegate`, `then`, `MyGeometry: struct:Geometry`, `InDeltaTime: float`).

The custom event compiles successfully; the BP runs; the user's chain wired off the custom event's `then` exec **never fires** because the inherited `UUserWidget::Tick` lifecycle hook isn't bound. No warning, no error.

Same failure shape applies to any AActor-aliased name typed against a non-AActor parent that declares the original (un-aliased) name — `BeginPlay`, `EndPlay`, `Hit`, `Destroyed`, etc., on UMG / `UAnimInstance` / `UGameInstanceSubsystem` / etc.

## Affected configurations

Default. No flags or settings gate this.

## Reproduction

1. Have a Widget Blueprint at `/Game/UI/WBP_Test` (or any `UUserWidget` subclass BP).
2. Remove the factory-spawned Tick override from the EventGraph (the factory auto-spawns one on widget BP creation; you need a clean graph to see the bug).
3. Call `add_event_node`:
   ```
   blueprint_query.add_event_node({
     asset_path: "/Game/UI/WBP_Test",
     event_name: "Tick"
   })
   ```
4. Inspect the response: `class: "CustomEvent"`, `is_override: false`. Expected: `class: "UserWidget"`, `is_override: true`.
5. (Optional confirmation) Wire the custom event's `then` exec to a `Print String` node, compile, save, run PIE. The string never prints — the inherited Tick was never bound.

## Diagnosis

UE 5.7 confirms the function name on `UUserWidget` is `Tick` (not `ReceiveTick`):

```cpp
// Engine/Source/Editor/UMGEditor/Private/WidgetBlueprintCompiler.cpp:1044
const UFunction* ReceiveTickEvent = FKismetCompilerUtilities::FindOverriddenImplementableEvent(
    GET_FUNCTION_NAME_CHECKED(UUserWidget, Tick),  // ← function name is "Tick"
    NewWidgetBlueprintClass);
```

The local C++ symbol is named `ReceiveTickEvent` by convention but the lookup is for `UUserWidget::Tick`. Misleading variable name, correct semantics.

Inside Monolith:

- `MonolithBlueprintNodeActions.cpp` alias table (~line 2500): `{TEXT("tick"), TEXT("ReceiveTick")}` (and similar for the AActor `ReceiveX` family).
- After alias resolution, the parent-chain walk uses `FName(ResolvedEventName)`, then `TestClass->FindFunctionByName(EventFName, EIncludeSuperFlag::ExcludeSuper)` against each ancestor.
- On a `UUserWidget` BP, the chain is `UUserWidget → UVisual → UWidget → UObject`. None of them declare `ReceiveTick`. The walk leaves `EventFunc == nullptr` and `DeclaringClass == nullptr`.
- Control flow falls into the `else` branch and constructs `K2Node_CustomEvent` named "Tick".

## Root cause

The alias table assumes every `BlueprintImplementableEvent` host follows AActor's `ReceiveX` naming convention. UMG, AnimInstance, GameInstanceSubsystem, and others use the bare name. There's no fallback for the case where the alias actually intercepted but the parent chain doesn't declare the resolved name.

## Severity

Silent feature-binding failure on the only authoring path Monolith offers for inherited overrides on non-AActor BlueprintImplementableEvent hosts. The user gets a compiling, runnable BP whose Tick chain never fires — the symptom only surfaces during PIE when expected per-frame behavior is missing.

Affects every UMG widget BP that needs a Tick override authored via Monolith, plus any future or current `BlueprintImplementableEvent` host whose declared name diverges from AActor's `ReceiveX` pattern.

## Manual workaround until the fix lands

Right-click the widget BP's EventGraph in the editor, choose "Add Event → Tick" (or "Add Event → Construct" / "Destruct" / etc., depending on what's needed). The editor's native path constructs the override correctly — only the Monolith MCP path is broken.

## Fix

Add a fallback re-walk with the original un-aliased `EventName` when the alias-resolved walk found no `UFunction` AND the alias actually changed the input name. On a hit, realign both `EventFName` and `ResolvedEventName` so the downstream override-uniqueness check (`FindOverrideForFunction`), `SetExternalMember` call, and response telemetry all reference the function that actually exists on the resolved `DeclaringClass`. Purely additive — existing AActor `ReceiveX` resolution, unknown-name custom-event fallthrough, and duplicate-override paths are unchanged. PR linked separately.

danielandric:
Fixes #47.

## Summary

Fixes a silent `add_event_node` failure: typing `event_name="Tick"` against a `UUserWidget` BP returns a `K2Node_CustomEvent` instead of the inherited `UUserWidget::Tick` override. The custom event compiles but never fires on widget tick.

Cause: the alias table maps `tick → ReceiveTick` (AActor convention) and the parent-chain walk uses the resolved name. UMG declares `Tick`, not `ReceiveTick`, so the walk misses and the action falls through to the custom-event branch. Mechanism + UE 5.7 citation in the commit message.

Fix: when the alias-resolved walk found no `UFunction` AND the alias actually changed the name, retry the walk with the original un-aliased `EventName`. On a hit, realign both `EventFName` and `ResolvedEventName` so the override-uniqueness check (`FindOverrideForFunction`), `SetExternalMember`, and response telemetry all reference the function that exists on the resolved `DeclaringClass`. Purely additive — existing AActor `ReceiveX` resolution, unknown-name custom-event fallthrough, and duplicate-override paths are unchanged.

## Test report

**Positive — UMG override now resolves.** Throwaway `UUserWidget` BP, removed factory-spawned Tick override, then `add_event_node({event_name: "Tick"})`. Response:

```
class=UserWidget
is_override=true
event_name="Tick"
pins=[OutputDelegate, then, MyGeometry (struct:Geometry), InDeltaTime (float)]
```

Pin-for-pin identical to UE's factory-spawned Tick on the same parent class.

**Override-uniqueness on UMG — both name forms realign.** Same `UUserWidget` BP with factory Tick override still present, `add_event_node({event_name: "Tick"})`. Response: error `Override event 'Tick' already exists`. Confirms `ResolvedEventName` was realigned to `"Tick"` (the message format is `*ResolvedEventName`) AND `FindOverrideForFunction(..., EventFName)` found the existing override using the realigned `EventFName="Tick"`.

**Negative — AActor unchanged.** Throwaway `AActor` BP with factory `ReceiveTick` override, `add_event_node({event_name: "Tick"})`. Response: error `Override event 'ReceiveTick' already exists`. First walk resolved `AActor::ReceiveTick` directly; fallback gate `!EventFunc` was false so the fallback was skipped.

## Residual risk surface

- Events outside the alias table (`Construct`, `Destruct`, `OnPaint`, `OnInitialized`, `BlueprintUpdateAnimation`, etc.) already worked: no alias intercept means the first walk finds them directly. The fix doesn't change their behavior either way.
- Cost: one extra `FindFunctionByName` walk per call, only when the alias changed the name AND the first walk missed. Bounded to the ~10 AActor-aliased names on non-AActor parents.
- The realignment changes the response's `event_name` from the alias to the original name on the new path. Consistent with the existing AActor path where `event_name` reflects the resolved `ReceiveX` — both report the function actually bound.

## Alternatives considered

- Extend the alias table with explicit UMG entries (`widget_tick → Tick`, etc.). Rejected: doesn't address bare-name input (`event_name: "Tick"`) which is what callers type, and adds maintenance burden for every new BlueprintImplementableEvent host class.
- Walk the parent chain once with a candidate-list `[ResolvedEventName, EventName]`. Rejected: larger blast radius for a one-call-cheap second walk; the explicit fallback is easier to reason about and revert if needed.

### DIFF
--- a/Source/MonolithBlueprint/Private/MonolithBlueprintNodeActions.cpp
+++ b/Source/MonolithBlueprint/Private/MonolithBlueprintNodeActions.cpp
@@ -2457,6 +2457,32 @@ FMonolithActionResult FMonolithBlueprintNodeActions::HandleAddEventNode(const TS
 		}
 	}
 
+	// Fallback: when the alias resolved to a name the parent chain doesn't
+	// declare (e.g. "Tick" -> "ReceiveTick" against UUserWidget, which declares
+	// "Tick" not "ReceiveTick"), retry with the original un-aliased EventName.
+	// Without this, callers typing AActor-style names on a non-AActor parent
+	// silently fall through to K2Node_CustomEvent and never bind to the
+	// inherited override — runtime is silent.
+	if (!EventFunc && ResolvedEventName != EventName && ParentClass)
+	{
+		FName OriginalFName(*EventName);
+		for (UClass* TestClass = ParentClass; TestClass; TestClass = TestClass->GetSuperClass())
+		{
+			UFunction* TestFunc = TestClass->FindFunctionByName(OriginalFName, EIncludeSuperFlag::ExcludeSuper);
+			if (TestFunc)
+			{
+				DeclaringClass = TestClass;
+				EventFunc = TestFunc;
+				// Realign both forms of the name — the override-uniqueness check uses
+				// EventFName, SetExternalMember below uses EventFName, and the response
+				// telemetry uses ResolvedEventName.
+				ResolvedEventName = EventName;
+				EventFName = OriginalFName;
+				// Keep walking up — match the resolved-name walk's topmost-declarer contract
+			}
+		}
+	}
+
 	// If we found a native event in the inheritance chain, create an override event node
 	if (DeclaringClass && EventFunc)
 	{


tumourlove:
Nice fix - tested with a UserWidget BP and it correctly creates a K2Node_Event override for Tick instead of falling through to CustomEvent. Closes #47. Will be in the next release (this weekend).

tumourlove:

Thanks for the widget-`Tick` fix, @danielandric! Closes Issue #47. The misleading C++ symbol name `ReceiveTickEvent` was bait — the real engine name is `Tick`, and your retry-with-un-aliased-name on lookup miss is the right call.

PR #48, opened 2026-04-29
feat(editor): run_automation_tests + list_automation_tests actions

MaxenceEpitech:
## Summary

Two new `editor` actions that let agents enumerate and run UE automation tests via MCP, **inside the running editor process** — no PIE, no commandlet, no second editor instance.

The existing alternative (`UnrealEditor.exe MyProject.uproject -ExecCmds=\"Automation RunTests <prefix>;Quit\" -unattended`) fails with `\"Existing instance is the same age or newer\"` while the dev editor is open (uproject file-lock). This action calls `FAutomationTestFramework::StartTestByName` + `StopTest` directly in-process, sidestepping that constraint.

## API

```json
// editor.list_automation_tests
{ "prefix": "MyProject.Weapon" }   // optional; empty = all tests
```

→ `{ "count": N, "tests": [{full_path, display_name, test_name, flags}] }`

```json
// editor.run_automation_tests
{ "prefix": "MyProject.WeaponX", "max_tests": 200 }
```

→
```json
{
  "prefix": "MyProject.WeaponX",
  "success": true,
  "total": 3, "passed": 3, "failed": 0, "skipped": 0,
  "results": [
    { "full_path": "...", "status": "passed", "duration_seconds": 0.04,
      "error_count": 0, "warning_count": 0 }
  ]
}
```

On failure each result also carries an `errors: [string]` array with the captured `FAutomationExecutionEntry` messages.

## Use case

Lock down regressions on a calibrated asset (e.g. a weapon's bow attach offsets, arrow scale, AnimBP wiring, character components) by writing `IMPLEMENT_SIMPLE_AUTOMATION_TEST` smoke tests that load the assets and assert frozen values, then drive the suite from an agent loop:

```
mcp__monolith__editor_query("run_automation_tests", {prefix: "MyProject.Weapon"})
```

## Notes

- **No new module dependency** — `FAutomationTestFramework` lives in Engine, already on the editor module's transitive deps.
- **Sync only** — latent / async tests (TickTests-driven) are not exercised by this code path and are reported with `status=\"skipped\"` so callers know they were not run. A follow-up could add a polling-based latent runner if needed.
- **`flags` field** is exposed as a number (cast from `EAutomationTestFlags` enum class via `uint32`) for forward compat with future enum additions.

## Test plan

- [ ] Build the plugin in a UE 5.7 project with at least one `IMPLEMENT_SIMPLE_AUTOMATION_TEST` registered
- [ ] `list_automation_tests` returns the test (with optional prefix filter)
- [ ] `run_automation_tests` with matching prefix returns `success=true, passed=N` for a passing suite
- [ ] Intentionally break a test assertion → verify `success=false, failed=1`, and the error message is in `results[0].errors`
- [ ] Edge: empty prefix match → returns `message` field, no crash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Added
+
+- **`editor.run_automation_tests` + `editor.list_automation_tests` actions** — Run / enumerate UE automation tests by full-path prefix (e.g. `MazeLegends.Bow`) via `FAutomationTestFramework` from inside the running editor. No PIE, no commandlet, no second editor process — sidesteps the `.uproject` file-lock that prevents `UnrealEditor -ExecCmds="Automation RunTests <prefix>"` from running while the editor is open. Returns a structured JSON summary (`success`, `total`, `passed`, `failed`, `skipped`) plus per-test results with error messages so agents can drive a regression suite end-to-end (e.g. "lock down a calibrated weapon's data-asset values; assert across edits"). Editor action count: 22 → 24. PR by **@MaxenceEpitech**.
+
 ## [0.14.7] - 2026-04-26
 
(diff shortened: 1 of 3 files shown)


MaxenceEpitech:
## Smoke + regression test on a real project ✅

Pushed two follow-up fixes after live-testing the action against a 3-suite test pack on a UE 5.7 game project:

**Bug 1 — empty result list.** UE's `RequestedTestFilter` defaults to `SmokeFilter` only. Game-module tests typically register with `ProductFilter`, so `GetValidTestNames()` returned engine tests (395) and zero project tests on a fresh editor session. Fix: widen the filter to all buckets before enumerating.

**Bug 2 — crash on test run.** `StartTestByName` keys the registry by **class name** (e.g. `FBowDataAssetTest`), not the human-readable full path. Passing the full path failed silently, left `GIsAutomationTesting=false`, and the next `StopTest` call tripped `check(GIsAutomationTesting)` → assertion / editor crash. Fix: use `Info.GetTestName()` as the lookup key, pass the readable path as the optional `InFullTestPath` arg so engine logs still show it, and gate the call on `ContainsTest()` so a stale or malformed entry returns `status=skipped` instead of crashing.

### Verified

\`\`\`
mcp.editor_query("run_automation_tests", {"prefix": "MazeLegends.Bow"})
→ {success:true, total:3, passed:3, failed:0, skipped:0}

# regression: bump DA_Bow.ArrowScale3P to (999,999,999) and re-run
→ {success:false, failed:1, results:[{
     full_path: "MazeLegends.Bow.DataAsset",
     status: "failed",
     errors: ["Expected 'ArrowScale3P uniform 1.15' to be 1.150000, but it was 999.000000 and outside tolerance 0.005000."]
   }, ...]}

# restore the value, re-run → success:true
\`\`\`

End-to-end MCP-driven CI loop now works without a second editor process.

tumourlove:
Thanks for this! Tested locally and it works great - 80 Monolith tests enumerated, ran a batch without issues. Clean structured output too. Will be included in the next release (this weekend).

tumourlove:

Thanks for the `editor.run_automation_tests` + `list_automation_tests` action pair, @MaxenceEpitech! Driving UE automation tests in-process via `FAutomationTestFramework` (no PIE / commandlet / second editor) sidesteps the `.uproject` file-lock entirely. The follow-up filter-widen + class-name-key + crash-guard hardening landed on top.

PR #51, opened 2026-05-02
feat(animation): add copy_bone_pose_between_sequences action

MaxenceEpitech:
## Summary
- Adds a new animation action `copy_bone_pose_between_sequences` that reads the **evaluated** bone transform (track + ref pose fallback) from a source `UAnimSequence` at a given time, then writes it as keys to a destination sequence for a list of bones.
- Closes a gap in the existing `get_bone_track_keys` / `set_bone_track_keys` flow: the read side fails with `Bone track not found` whenever a bone has no explicit track in the source FBX (typical when an animator only keyed bones that move). The new action sidesteps this by using `UAnimSequence::GetBoneTransform(..., bUseRawData=true)`, which falls back to the skeleton's ref pose for unkeyed bones.

## Use cases
- Repair a partial T-pose: a sword idle whose left arm is in T-pose can have just the left-arm chain copied from a working `Walk_1` frame 0 — without touching the right arm / sword grip / spine pose.
- Re-pose / patch an existing anim without re-authoring the source FBX.
- Bake a static pose from any frame of an animated sequence (just pass `apply_to_all_dest_frames=true`, default).

## API
```
copy_bone_pose_between_sequences:
  source_path: string  (required) — source AnimSequence
  dest_path: string    (required) — destination AnimSequence
  bone_names: string[] (required) — bones to copy
  source_time: number  (optional, default 0.0) — time on source to evaluate
  apply_to_all_dest_frames: bool (optional, default true) — write to every dest frame (static pose) or only frame 0
```
Returns `copied_bones`, `skipped_bones` (with reason), `keys_written_per_bone`, and the original params for traceability.

## Test plan
- [ ] Call with a known-good `Walk_1.uasset` as source and a broken `Idle.uasset` as dest, copying `["upperarm_l","lowerarm_l","hand_l"]` at `source_time=0.0`. Verify the dest's left arm now matches Walk's authored frame-0 pose, while right-arm / spine / leg bones are unchanged.
- [ ] Call with a bone that doesn't exist in either skeleton — confirm it lands in `skipped_bones` with the right `reason`.
- [ ] `apply_to_all_dest_frames=false` writes only frame 0 (subsequent frames keep their previous data).
- [ ] Open dest in UE5 Anim Editor — keys appear correctly on the timeline; pose preview matches expectation.
- [ ] Confirm the action is reachable both directly and via `batch_execute`.

### DIFF
--- a/Source/MonolithAnimation/Private/MonolithAnimationActions.cpp
+++ b/Source/MonolithAnimation/Private/MonolithAnimationActions.cpp
@@ -226,6 +226,17 @@ void FMonolithAnimationActions::RegisterActions(FMonolithToolRegistry& Registry)
 			.Optional(TEXT("include_children"), TEXT("bool"), TEXT("Also remove child bone tracks"), TEXT("false"))
 			.Build());
 
+	Registry.RegisterAction(TEXT("animation"), TEXT("copy_bone_pose_between_sequences"),
+		TEXT("Copy evaluated bone transforms (track + ref pose fallback) from a source AnimSequence at a given time to a destination AnimSequence as keys"),
+		FMonolithActionHandler::CreateStatic(&HandleCopyBonePoseBetweenSequences),
+		FParamSchemaBuilder()
+			.Required(TEXT("source_path"), TEXT("string"), TEXT("Source AnimSequence asset path"))
+			.Required(TEXT("dest_path"), TEXT("string"), TEXT("Destination AnimSequence asset path"))
+			.Required(TEXT("bone_names"), TEXT("array"), TEXT("Array of bone names to copy"))
+			.Optional(TEXT("source_time"), TEXT("number"), TEXT("Time in seconds on source to evaluate (default 0.0)"), TEXT("0.0"))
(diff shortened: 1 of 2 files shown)


tumourlove:
Cherry-picked into master at `a547996` (your authorship preserved) and shipped in [v0.14.9](https://github.com/tumourlove/monolith/releases/tag/v0.14.9), with three small maintainer follow-ups landed as a separate commit on top: (1) replaced the UE 5.6-deprecated `GetBoneTransform(FTransform&, FSkeletonPoseBoneIndex, double, bool)` overload with the non-deprecated `FAnimExtractContext(SourceTime)` form (drops the `PRAGMA_DISABLE_DEPRECATION_WARNINGS` shim), (2) added `bone_names` array element-type guard returning `-32602` on non-string entries, (3) added `SourceTime` clamp to `[0, GetPlayLength()]` with `original_source_time` + `clamped_source_time` surfaced in the response. Thanks @MaxenceEpitech for closing this workflow gap.

tumourlove:
Shipped in v0.14.9 via manual cherry-pick (your authorship preserved on the action commit). Maintainer follow-up landed as a separate commit on top. Closing the PR; the work is in master. Thanks!

PR #54, opened 2026-05-03
fix(animation) + feat(editor): get_bone_track_keys deprecated-API fix, list_bone_tracks helper, run_console_command + start/stop_pie

MaxenceEpitech:
## Summary

Two unrelated-but-bundled changes I hit while wiring an external test rig against the editor:

### `fix(animation): get_bone_track_keys` — drop deprecated `GetBoneAnimationTracks()`

`HandleGetBoneTrackKeys` was reading the raw `FRawAnimSequenceTrack` via the deprecated `IAnimationDataModel::GetBoneAnimationTracks()` accessor. That path returns the **uncompressed source tracks**, which are missing on AnimSequences that have already been baked / compressed — so callers got `Bone track not found: <bone>` even when the bone was clearly animated and visible in the asset.

Switched to the public, non-deprecated API:
- `IAnimationDataModel::IsValidBoneTrackName()` to validate the bone
- `IAnimationDataModel::GetBoneTrackTransforms()` to evaluate per-key `FTransform`s

Works regardless of underlying compressed storage. Also gives us proper scale keys (the old code silently dropped scales when key counts differed across pos/rot/scale arrays).

Adds an `AllTransforms.Num() == 0` guard so an empty track returns a clean error instead of producing `num_keys=0` and a misleading `start_frame > end_frame` message.

### `feat(animation): add `list_bone_tracks` helper

Returns every bone name that has a track in a given `UAnimSequence`. You need this to drive `get_bone_track_keys` without guessing names — Skeleton bone listings include unanimated bones, so they're not a substitute. Wired into `batch_execute` alongside the other animation read actions.

### `feat(editor): add `run_console_command` + `start_pie` / `stop_pie`

Three small actions that let external tooling drive in-editor / PIE testing without keyboard input:

- `editor.run_console_command` — dispatches a console command via the first `APlayerController` of the active PIE world (so exec UFUNCTIONs on the possessed pawn fire correctly), falling back to `GEngine->Exec` on the editor world when no PC is available. Returns which world type was used and whether the PC path was taken.
- `editor.start_pie` — queues a default Play-In-Editor session via `GEditor->RequestPlaySession` + `StartQueuedPlaySessionRequest`. Refuses to queue a duplicate when a PIE world is already alive.
- `editor.stop_pie` — `RequestEndPlayMap` when a PIE world exists, no-op (with `stopped=false`) otherwise.

Pairs naturally with the existing `run_python` / `load_level` (HOFF 7) actions for fully automated in-game test flows: load level → start PIE → run console cmds → stop.

## Test plan
- [ ] Build plugin against UE 5.7 (Mac); confirm both modules compile
- [ ] `animation.list_bone_tracks` on a compressed AnimSequence returns non-empty `bone_names`
- [ ] `animation.get_bone_track_keys` on a compressed AnimSequence returns `num_keys > 0` for an animated bone (was failing before this PR with "Bone track not found")
- [ ] `editor.start_pie` spawns PIE; `editor.run_console_command` with an exec UFUNCTION on the player pawn fires it; `editor.stop_pie` cleanly tears down

## Notes
- Two commits, kept separate so animation vs editor land independently if you'd rather split the PR.
- Branched off `master` at 831fdac (post v0.14.9 docs refresh).

### DIFF
--- a/Source/MonolithAnimation/Private/MonolithAnimationActions.cpp
+++ b/Source/MonolithAnimation/Private/MonolithAnimationActions.cpp
@@ -291,6 +291,12 @@ void FMonolithAnimationActions::RegisterActions(FMonolithToolRegistry& Registry)
 			.Optional(TEXT("start_frame"), TEXT("integer"), TEXT("Start frame (default 0)"), TEXT("0"))
 			.Optional(TEXT("end_frame"), TEXT("integer"), TEXT("End frame (default -1 = all)"), TEXT("-1"))
 			.Build());
+	Registry.RegisterAction(TEXT("animation"), TEXT("list_bone_tracks"),
+		TEXT("List all bone names that have tracks (animated bones) in an AnimSequence"),
+		FMonolithActionHandler::CreateStatic(&HandleListBoneTracks),
+		FParamSchemaBuilder()
+			.Required(TEXT("asset_path"), TEXT("string"), TEXT("AnimSequence asset path"))
+			.Build());
 	Registry.RegisterAction(TEXT("animation"), TEXT("get_sequence_curves"),
 		TEXT("Get float and transform curves on an animation sequence"),
 		FMonolithActionHandler::CreateStatic(&HandleGetSequenceCurves),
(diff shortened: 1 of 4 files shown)


tumourlove:
Thanks @MaxenceEpitech! This is in master since commits 5bc5a75 / cfd0bff / 39eb313 and credited in the v0.14.10 [Unreleased] CHANGELOG section. Will ship in the v0.14.10 release. Closing.

PR #56, opened 2026-05-07
feat(animation): add {get,add,remove}_compatible_skeleton actions

MaxenceEpitech:
## Summary

Wraps `USkeleton::CompatibleSkeletons` via three new actions on the `animation` namespace.

This is the canonical UE5 mechanism that lets anims authored on one skeleton play on another (the typical case: UE4 mannequin animation packs on UE5 `SK_Mannequin` characters). Previously the only way to declare or inspect this from Monolith was via `editor_query.run_python` + manual `unreal.Skeleton.add_compatible_skeleton` boilerplate.

### `animation.get_compatible_skeletons`

```
{ asset_path: string }   →   { asset_path, compatible_skeletons: [string], count }
```

Lists the current `CompatibleSkeletons` entries on a `USkeleton`.

### `animation.add_compatible_skeleton`

```
{ asset_path: string, compatible_with: string, save?: bool=true }
  →
{ asset_path, compatible_with, added: bool, already_compatible: bool, count }
```

Idempotent — skips if the target skeleton is already in the list. Marks the package dirty + saves by default. Errors on self-compat or unresolved target.

### `animation.remove_compatible_skeleton`

Symmetric removal:

```
{ asset_path: string, compatible_with: string, save?: bool=true }
  →
{ asset_path, compatible_with, removed: bool, was_compatible: bool, count }
```

## Why these matter

In a real session integrating a Marketplace UE4 spear animation pack onto a UE5 first-person `SK_Mannequin` character, the only reason animation playback worked at all was that the original asset import had pre-populated the compatibility list. Diagnosing whether that was the case + flipping it on demand required dropping into `run_python`. With these actions:

```
mcp__monolith__animation_query get_compatible_skeletons asset_path=/Game/.../SK_Mannequin
mcp__monolith__animation_query add_compatible_skeleton \
    asset_path=/Game/.../SK_Mannequin \
    compatible_with=/Game/.../UE4_Mannequin_Skeleton
```

becomes the single-line replacement.

## Pattern conformance

Matches the surrounding animation actions: `FParamSchemaBuilder` registration block placed alongside `get_abp_*`, `FMonolithAssetUtils::LoadAssetByPath` for both inputs, JSON envelope with explicit `count` and disjoint booleans (`added` / `already_compatible`, `removed` / `was_compatible`) so callers can distinguish no-op from change.

No new module dependencies (`UEditorAssetLibrary` + `USkeleton::AddCompatibleSkeleton/RemoveCompatibleSkeleton/GetCompatibleSkeletons` are already in the existing build deps).

## Test plan

- [ ] Compile clean on UE 5.7
- [ ] `get_compatible_skeletons` on a fresh skeleton returns `count: 0`
- [ ] `add_compatible_skeleton A B` then `get_compatible_skeletons A` returns `[B]`, `count: 1`
- [ ] Calling `add_compatible_skeleton A B` again returns `already_compatible: true, added: false, count: 1`
- [ ] `add_compatible_skeleton A A` returns an error (self-compat rejected)
- [ ] `remove_compatible_skeleton A B` returns `removed: true`, then `count: 0`
- [ ] `remove_compatible_skeleton A B` again returns `was_compatible: false, removed: false`
- [ ] With `save: false`, the asset stays dirty after the action

### DIFF
--- a/Source/MonolithAnimation/Private/MonolithAnimationActions.cpp
+++ b/Source/MonolithAnimation/Private/MonolithAnimationActions.cpp
@@ -649,6 +649,30 @@ void FMonolithAnimationActions::RegisterActions(FMonolithToolRegistry& Registry)
 			.Required(TEXT("asset_path"), TEXT("string"), TEXT("Animation Blueprint asset path"))
 			.Build());
 
+	// Skeleton Compatibility — required for legacy UE4 anims on UE5 mannequin etc.
+	Registry.RegisterAction(TEXT("animation"), TEXT("get_compatible_skeletons"),
+		TEXT("List skeletons declared compatible with the given Skeleton (USkeleton::CompatibleSkeletons)"),
+		FMonolithActionHandler::CreateStatic(&HandleGetCompatibleSkeletons),
+		FParamSchemaBuilder()
+			.Required(TEXT("asset_path"), TEXT("string"), TEXT("Skeleton asset path"))
+			.Build());
+	Registry.RegisterAction(TEXT("animation"), TEXT("add_compatible_skeleton"),
+		TEXT("Declare another Skeleton as compatible — lets anims authored on the other skeleton play on this one without retargeting"),
+		FMonolithActionHandler::CreateStatic(&HandleAddCompatibleSkeleton),
+		FParamSchemaBuilder()
+			.Required(TEXT("asset_path"), TEXT("string"), TEXT("Skeleton asset path (the one being declared compatible)"))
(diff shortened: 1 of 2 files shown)


tumourlove:
Thanks @MaxenceEpitech! Cherry-picked to master via 91d9e57. Smoke-tested all three actions against duplicated SK_Mannequin skeletons -- empty-get/add/re-add idempotency, self-compat rejection, get-after-add, remove/remove-again disjoint-return, save flag -- all green. Closes the cross-skeleton retarget setup workflow nicely. Will ship in the v0.14.10 release. Closing.

PR #57, opened 2026-05-07
feat(blueprint): add owner_class_filter / name_pattern / exclude_categories to get_cdo_properties

MaxenceEpitech:
## Summary

Three new optional filters on `blueprint.get_cdo_properties` that make the output usable on `AActor` / `APawn` / `ACharacter` subclasses without wading through 200+ inherited engine properties on every call.

### `owner_class_filter` (string)

Case-insensitive substring match on the property's owning class name. Pass `"MazeLegends"` and the response only contains props declared on `MazeLegendsCharacter` (and any further subclass). Skips everything inherited from `AActor` / `APawn` / `ACharacter` in one parameter — no need to enumerate categories or fight the inheritance chain.

### `name_pattern` (string)

Case-insensitive substring match on the property name. One round-trip to find every "Weapon*", "Camera*", "Mesh*" property without grepping the full payload.

### `exclude_categories` (array of strings)

Case-insensitive exact match on the property's `Category` metadata. One pass removes the noisiest engine categories:

```
exclude_categories: ["Replication", "Cooking", "HLOD", "Lighting", "Rendering"]
```

## Why

In a recent session inspecting `BP_FirstPersonCharacter`, `get_cdo_properties` returned 202 properties — most of them generic `AActor` flags that aren't relevant when you're trying to read project-specific config. Filtering server-side cuts the JSON payload by ~90% in typical use and saves a meaningful amount of context tokens for the LLM caller.

## Compatibility

All three params are additive and default to "no filtering". Existing callers see no behavior change. They compose with the pre-existing `category_filter` and `include_parent_defaults` options:

```
asset_path: /Game/.../BP_FirstPersonCharacter
include_parent_defaults: true
owner_class_filter: "MazeLegends"
name_pattern: "Weapon"
exclude_categories: ["Replication"]
```

## Test plan

- [ ] Compile clean on UE 5.7
- [ ] No filters → same output as before this PR
- [ ] `owner_class_filter: "MazeLegends"` on an `AMazeLegendsCharacter` BP returns only props declared on `MazeLegendsCharacter`
- [ ] `name_pattern: "weapon"` returns only properties whose name contains "weapon" (case-insensitive)
- [ ] `exclude_categories: ["Replication"]` removes all properties whose `Category` metadata equals "Replication"
- [ ] All three combined still respects `category_filter` and `include_parent_defaults`

### DIFF
--- a/Source/MonolithBlueprint/Private/MonolithBlueprintCDOActions.cpp
+++ b/Source/MonolithBlueprint/Private/MonolithBlueprintCDOActions.cpp
@@ -26,6 +26,9 @@ void FMonolithBlueprintCDOActions::RegisterActions(FMonolithToolRegistry& Regist
 			.Required(TEXT("asset_path"), TEXT("string"), TEXT("Asset path (e.g. /Game/Blueprints/BP_MyActor or /Game/Data/DA_MyData)"))
 			.Optional(TEXT("category_filter"), TEXT("string"), TEXT("Only include properties whose category contains this string"))
 			.Optional(TEXT("include_parent_defaults"), TEXT("boolean"), TEXT("If true, include properties inherited from native parent class (default: true)"))
+			.Optional(TEXT("owner_class_filter"), TEXT("string"), TEXT("Only include properties whose owner_class name contains this string (case-insensitive). Lets you skip everything inherited from AActor/APawn/ACharacter when only project-level props matter."))
+			.Optional(TEXT("name_pattern"), TEXT("string"), TEXT("Only include properties whose name contains this substring (case-insensitive)"))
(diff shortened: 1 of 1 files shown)


tumourlove:
Thanks @MaxenceEpitech! Cherry-picked to master via 3e13d9b. Smoke-tested all three new params against SandboxCharacter_CMC -- owner_class_filter strips the inherited Character/Pawn/Actor surface (197 -> 61 props), name_pattern filters case-insensitively, exclude_categories works as exact-match-array, and the four-filter composition yields 1 property from 197 (99.5% payload reduction). All filters compose with the pre-existing category_filter + include_parent_defaults. Will ship in the v0.14.10 release. Closing.

Issue #59, opened 2026-05-16
Niagara summary APIs need clearer guidance and richer semantics for event-linked emitter systems


middle233:
## Summary

  When using Monolith's Niagara MCP actions on multi-emitter systems, the AI can misinterpret emitter relationships if the system uses Niagara events
  to connect emitters.

  A concrete failure mode is a firework-style setup:
  - one emitter acts as the shell / leader
  - one emitter follows that shell via a location event
  - one emitter is intended to burst when the shell reaches a meaningful point or dies

  Without explicit semantic guidance, the AI may treat the burst emitter as an independent emitter and leave it spawning at its own local origin
  instead of reasoning about whether it should be driven by an inter-emitter event chain.

  ## Problem

  The raw structural data exposed by the existing Niagara summary/read APIs is not always enough for an LLM to reliably infer:
  - which emitter is an event source
  - which emitter is an event receiver
  - whether a burst-like emitter is independent or event-driven
  - whether spawn placement is local-shape-based or event-payload-based
  - whether `requires_persistent_ids` may matter for a cross-emitter event workflow

  This is especially easy to misread when the system contains:
  - `GenerateLocationEvent` / `GenerateDeathEvent`
  - event handlers on another emitter
  - local location modules such as `ShapeLocation`
  - emitter names like `Shell`, `Trail`, `Burst`, `Explosion`, etc.

  ## Proposed improvement

  Niagara summary APIs should expose a higher-level semantic layer in addition to raw structure.

  Useful semantic fields include:
  - inter-emitter event topology
  - generated events per emitter
  - consumed events per emitter
  - inferred `spawn_location_mode`
  - inferred `role_hint`
  - `requires_persistent_ids`
  - warnings/suggestions in validation when a burst-like emitter appears independent but likely should be event-driven

  It is also important to keep default summary output compact for token efficiency.

  ## Suggested API behavior

  For `niagara.get_system_summary` and `niagara.get_emitter_summary`:
  - default to `detail_level=compact`
  - support `detail_level=full` for deeper inspection
  - recommend `full` when emitters may be linked by Niagara events

  For `niagara.validate_system`:
  - add checks for unresolved or mismatched event chains
  - flag burst/explosion-like emitters that have no incoming event handler and therefore behave independently
  - surface `requires_persistent_ids` guidance when event-linked emitter workflows are detected

  ## Why this matters

[...]

PR #60, opened 2026-05-16
Improve Niagara summary semantics for event-linked emitter workflows

middle233:
## Summary

  This PR improves the Niagara MCP read/validation surface for multi-emitter systems that use Niagara events.

  It adds higher-level semantic analysis so AI callers can better reason about:
  - event source vs event receiver emitters
  - local-shape spawning vs event-driven spawning
  - burst-like emitters that are actually independent
  - when `requires_persistent_ids` may matter

  ## Changes

  - Enhanced `niagara.get_system_summary`
  - Enhanced `niagara.get_emitter_summary`
  - Added `detail_level=compact|full` to both summary actions
  - Kept `compact` as the default to limit token usage
  - Expanded `niagara.validate_system` with event-chain and persistent-id guidance
  - Updated action descriptions to recommend `full` when emitters may be linked by Niagara events

  ## Notes

  This change is about improving the MCP surface, not authoring a specific Niagara asset.

  Only Niagara plugin code was changed:
  - `Source/MonolithNiagara/Private/MonolithNiagaraActions.cpp`

  ## Validation

  Built successfully with:
  - `MyProjectEditor Win64 Development`

### DIFF
--- a/Source/MonolithNiagara/Private/MonolithNiagaraActions.cpp
+++ b/Source/MonolithNiagara/Private/MonolithNiagaraActions.cpp
@@ -704,6 +704,600 @@ static FString GetAssetPath(const TSharedPtr<FJsonObject>& Params)
 	return Path;
 }
 
+namespace
+{
+	enum class EMonolithSemanticDetailLevel
+	{
+		Compact,
+		Full
+	};
+
+	static bool TryParseSemanticDetailLevel(
+		const TSharedPtr<FJsonObject>& Params,
+		EMonolithSemanticDetailLevel& OutDetailLevel,
+		FString& OutError)
+	{
+		const FString DetailLevel = Params->HasField(TEXT("detail_level"))
+			? Params->GetStringField(TEXT("detail_level")).ToLower()
+			: TEXT("compact");
+
+		if (DetailLevel.IsEmpty() || DetailLevel == TEXT("compact"))
+		{
+			OutDetailLevel = EMonolithSemanticDetailLevel::Compact;
+			return true;
+		}
+
+		if (DetailLevel == TEXT("full"))
+		{
+			OutDetailLevel = EMonolithSemanticDetailLevel::Full;
+			return true;
+		}
+
+		OutError = FString::Printf(TEXT("Invalid detail_level '%s'. Valid values: compact, full"), *DetailLevel);
+		return false;
+	}
+
+	struct FMonolithNiagaraStageModule
+	{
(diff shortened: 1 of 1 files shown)


tumourlove:
Thank you @middle233 — integrated into master locally. Will ship in the next release (v0.14.11). This also closes #59 (which you filed describing the same problem).

The PR's semantic layer landed in full: `detail_level=compact|full` on `get_system_summary` / `get_emitter_summary`, the event-chain + persistent-id rules on `validate_system`, the `AnalyzeEmitterSemantic` classifier with role hints / spawn-location modes / topology fields. All three of your example failure modes (firework shell+trail+burst, death-event bursts, location-event followers) are now reasoned about by the new semantic layer.

Two small maintainer changes on top of the PR:

1. **Performance:** `HandleValidateSystem` invoked `AnalyzeEmitterSemantic` once per emitter in the outer loop AND again inside the inner source-emitter lookup — O(N²) graph walks on many-emitter event-heavy systems. Cached results in a `TMap<FGuid, FMonolithNiagaraEmitterSemantic>` built once before the validation loop; inner loop now does `TMap::Find(SourceGuid)`. Same shape, just doesn't re-walk graphs.

2. **Cleanup:** `GetGraphForHandleUsage`'s `SystemSpawn` / `SystemUpdate` switch case was dead — the only caller (`CollectEmitterModules`) only ever passes emitter-stage usage. Dropped the branch.

One thing worth flagging in the CHANGELOG (and which I've noted there): the legacy `event_handlers[]` array on `get_emitter_summary` is removed by this PR — any external caller reading `event_handlers[].source_emitter_id` migrates to `incoming_events[].source_emitter_id` (`full` only) or `consumed_events[]`. Verified no internal callers depended on it.

Smoke-tested live on Epic's `NS_MuzzleFlash` and KINEMATION `NS_Casing_Rifle` — both are independent-emitter burst systems (no inter-emitter events) so they exercise the classifier but not the event-chain path; the new `semantic_notes` correctly flag local-shape spawning. The event-chain path is verified structurally via the source index but not yet against a live event-linked asset in this project. If you have a sample shell+trail+burst system you've used for testing, I'd be happy to run it as a follow-up validation.

Closing as integrated rather than merged because the commit landed under the maintainer handle to keep the release lineage linear — your contribution is credited in `CHANGELOG.md` and the release notes when v0.14.11 ships. Thanks for the well-scoped issue → PR pairing.

Issue #70, opened 2026-06-12
Crashing on 16k+ indexed assets


aggitti:
```

[2026.06.12-14.06.20:786][712]LogMonolithIndex: Deep indexed 16160 / 21640 assets (16160 ok, 0 errors)
[2026.06.12-14.06.20:793][712]LogStreaming: Display: FlushAsyncLoading(7715): 1 QueuedPackages, 0 AsyncPackages
[2026.06.12-14.06.20:795][712]LogStreaming: Warning: LoadPackage: SkipPackage: /Game/Blueprints/Core/BaseActors/BP_Base_UseAction (0x245BD8D9FCE09294) - The package to load does not exist on disk or in the loader
[2026.06.12-14.06.20:795][712]LogUObjectGlobals: Warning: Failed to find object 'Object /Game/Blueprints/Core/BaseActors/BP_Base_UseAction.BP_Base_UseAction_C'
[2026.06.12-14.06.20:795][712]LogOutputDevice: Warning: 

Script Stack (0 frames) :

[2026.06.12-14.06.20:806][712]LogWindows: Error: appError called: Assertion failed: MetaClass [File:D:\build\++UE5\Sync\Engine\Source\Runtime\CoreUObject\Private\UObject\PropertyClass.cpp] [Line: 160] 

Assertion failed: MetaClass [File:D:\build\++UE5\Sync\Engine\Source\Runtime\CoreUObject\Private\UObject\PropertyClass.cpp] [Line: 160] 

UnrealEditor_Core
UnrealEditor_CoreUObject
UnrealEditor_MonolithIndex!FUserDefinedStructIndexer::IndexAsset() [D:\Development\Unreal Projects\Game1\Dev\Game1\Plugins\Monolith\Source\MonolithIndex\Private\Indexers\UserDefinedStructIndexer.cpp:25]
UnrealEditor_MonolithIndex!`UMonolithIndexSubsystem::FIndexingTask::Run'::`215'::<lambda_7>::operator()() [D:\Development\Unreal Projects\Game1\Dev\Game1\Plugins\Monolith\Source\MonolithIndex\Private\MonolithIndexSubsystem.cpp:784]
[...]

PR #77, opened 2026-06-21
fix(mesh): gate GeometryScripting delay-load DLLs to Win64 (fixes Unique-env / macOS link)

itismyfield:
## Gate GeometryScripting delay-load DLLs to Win64

`MonolithMesh.Build.cs` adds `PublicDelayLoadDLLs.Add("UnrealEditor-GeometryScriptingCore.dll")` (+ GeometryFramework/GeometryCore) with the **hardcoded `UnrealEditor-` prefix**.

That prefix is only correct in a **Shared** build environment. A project using `TargetBuildEnvironment.Unique` (required when it changes engine-affecting compile settings) recompiles engine modules under the **project target prefix** (e.g. `CookingHeartEditor-`), so the hardcoded name no longer resolves.

On macOS this fails the link:
```
ld: library 'UnrealEditor-GeometryScriptingCore.dll' not found
clang++: error: linker command failed with exit code 1
```

### Fix
Gate the three `PublicDelayLoadDLLs.Add(...)` calls behind `Target.Platform == UnrealTargetPlatform.Win64`.

- Delay-load is a Windows PE mechanism; the `#70` `GetLastError=126` race it mitigates is Windows-specific.
- On non-Windows, the `PrivateDependencyModuleNames` (GeometryScriptingCore/Framework/Core, added just above) already link with the correctly-decorated prefix — no delay-load needed.
- No behavior change for Windows shared builds.

### Repro
Source-tree build, `BuildEnvironment.Unique`, GeometryScripting enabled, macOS, UE 5.8, Monolith v0.20.3 → MonolithMesh link fails as above. With this patch it links clean and Tier-5 mesh ops remain available (via the proper module dependency).

### DIFF
--- a/Source/MonolithMesh/MonolithMesh.Build.cs
+++ b/Source/MonolithMesh/MonolithMesh.Build.cs
@@ -116,9 +116,20 @@ public class MonolithMesh : ModuleRules
 			// removing the CouldNotBeLoadedByOS / GetLastError=126 first-build/load race).
 			// Verified: PublicDelayLoadDLLs at ModuleRules.cs:1308 (UE 5.7). Confirm via
 			// dumpbin that these imports move to the Delay Import table.
-			PublicDelayLoadDLLs.Add("UnrealEditor-GeometryScriptingCore.dll");
-			PublicDelayLoadDLLs.Add("UnrealEditor-GeometryFramework.dll");
-			PublicDelayLoadDLLs.Add("UnrealEditor-GeometryCore.dll");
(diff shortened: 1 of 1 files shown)


tumourlove:
Thanks for this — clean fix, and a good catch on the Unique-env / macOS prefix breakage. Gating the delay-loads to Win64 is exactly right since the PrivateDependencyModuleNames already cover the other platforms. Merging for the next release (v0.21.0).

PR #79, opened 2026-06-26
Fix dereferencing of Pair.Key in Remap addition

dulanw:
latest fails to build with 
`0>MonolithChooserActions.cpp(547,15): Error C2664 : 'FString FMonolithChooserActions::HandleDuplicateChooserTree::<lambda_1>::operator ()(const FString &) const': cannot convert argument 1 from 'const T' to 'const FString &'`


tumourlove:
Superseded by the fix in v0.21.0. I went with the `MonolithKeyToString` shim so both call sites are covered (including the second site in MonolithSkeletonRetargetActions) and it stays consistent with how the rest of the codebase reads JSON keys — but your diagnosis was correct and I appreciate the PR. You're credited in the release notes. Thanks @dulanw.

PR #102, opened 2026-07-23
Blueprint: inherited-SCS component overrides (ICH) + graph-node property actions

Thomasbehan:
## Summary

Two additions to the `MonolithBlueprint` module that make Blueprint automation work on component-based frameworks (RPG kits, GAS templates, etc.) where key data lives on **inherited** components and on **graph sub-object nodes**.

### 1. `set_component_property` now writes **inherited-SCS component** properties via an ICH override
Previously `set_component_property` resolved only (a) components added in *this* Blueprint's SimpleConstructionScript and (b) native inherited components (CDO default subobjects). A component declared in a **parent Blueprint's** SCS (e.g. a `StatsComponent`/`EquipmentComponent` defined several Blueprints up the chain) is `None` on the child CDO and absent from the child SCS, so the call failed with `Component not found`.

Now, when those two lookups miss, it walks the parent-BP chain for the defining `USCS_Node`, builds an `FComponentKey`, and **get-or-creates an Inheritable Component Handler (ICH) override template** for that key, then writes the property to it (structural-modify + recompile so the override persists). The response reports `source` = `scs` | `cdo_native` | `ich_override` so callers can see how it resolved.

This is the reflection-only equivalent of ticking an inherited component's property in the child's Details panel — now scriptable.

### 2. New actions: `set_graph_node_property` / `get_graph_node_properties`
Set/read a `UPROPERTY` on a graph **sub-object node** by object path (e.g. authoring Mutable `CustomizableObject` graph nodes, toggling a node's `EnabledState`, etc.). `set` optionally calls `ReconstructNode()` and saves. Uses `FProperty` reflection + public bases only (no `/Private/` engine headers), consistent with the existing CDO/component actions.

## Why
Component-based marketplace frameworks put the important knobs (health, startup loadout, etc.) on components inherited from deep base Blueprints. Without ICH support, none of it is scriptable from MCP. The graph-node action similarly unlocks sub-object graphs (Mutable, etc.) that the existing node actions don't reach.

## Testing
Built against **UE 5.7** (`WWJEditor Win64 Development`, clean compile). Verified live:
- `set_component_property` on a 3-levels-inherited `EquipmentComponent` → returns `source: ich_override`, the override persists across recompile, and the value is read back correctly (used it to populate `StartupWeapon`/`StartupInventory` `TArray<UserDefinedStruct>` properties).
- `set_graph_node_property` toggling `EnabledState` and setting struct properties on Mutable CO nodes.

## Notes
- No changes to existing behavior — the ICH path is a fallback only taken when the prior two lookups miss.
- Private engine classes avoided throughout; reflection + `UEdGraphNode::ReconstructNode()` / `FBlueprintEditorUtils` only.

### DIFF
--- a/Source/MonolithBlueprint/Private/MonolithBlueprintComponentActions.cpp
+++ b/Source/MonolithBlueprint/Private/MonolithBlueprintComponentActions.cpp
@@ -6,6 +6,8 @@
 #include "Engine/SimpleConstructionScript.h"
 #include "Engine/SCS_Node.h"
 #include "Engine/SkinnedAsset.h"
+#include "Engine/BlueprintGeneratedClass.h"
+#include "Engine/InheritableComponentHandler.h"
 #include "Components/SkinnedMeshComponent.h"
 #include "Kismet2/BlueprintEditorUtils.h"
 #include "Kismet2/KismetEditorUtilities.h"
@@ -56,7 +58,7 @@ void FMonolithBlueprintComponentActions::RegisterActions(FMonolithToolRegistry& 
 			.Build());
 
 	Registry.RegisterAction(TEXT("blueprint"), TEXT("set_component_property"),
-		TEXT("Set a property on a component template in a Blueprint via text import."),
+		TEXT("Set a property on a component template in a Blueprint via text import. Resolves BP-added SCS components, native CDO subobjects, AND SCS components inherited from a parent Blueprint (auto-creates an Inheritable Component Handler override, keyed by the defining ancestor SCS node). Reports 'source' = scs | cdo_native | ich_override."),
 		FMonolithActionHandler::CreateStatic(&HandleSetComponentProperty),
 		FParamSchemaBuilder()
 			.RequiredAssetPath(TEXT("asset_path"), TEXT("Blueprint asset path"))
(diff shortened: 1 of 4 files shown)


tumourlove:
Splitting this one — the two halves are getting different answers.

Part (a), the inherited-SCS component override, is queued for an upcoming release. The gap is real: `set_component_property` resolved SCS and native-CDO components only, so anything declared in a parent Blueprint's SCS failed outright. Verified the approach is sound — `CreateOverridenComponentTemplate` returns the existing template on a key match, so it does not clobber unrelated overrides, and the parent walk terminates.

Two things I will add when it lands. The engine guards this with `Key.IsValid() && BP->ParentClass && BP->ParentClass->IsChildOf(Key.GetComponentOwner())` (see `SubobjectData.cpp:264`, which is the 5.7-current path), and the PR omits it. It also needs an `FScopedTransaction` with `BP->Modify()` before the ICH mutation rather than after, or the override record is not undoable.

Part (b), the graph-node property actions, needs a narrower design before I can take it. The name undersells the surface: `ResolveNodeObject` accepts any object path and the write path never actually requires a `UEdGraphNode`, so it is effectively a set-any-UPROPERTY-on-any-UObject action that bypasses every allowlist in the module. Specifically: `save` defaults true and writes a `.uasset`, with `GetAssetPackageExtension()` hardcoding that extension so a node living in a level gets the wrong filename; `reconstruct` defaults true and `ReconstructNode()` can drop pin links; there is no `CPF_Edit`/`CPF_EditConst` filter on the write path, so `NodeGuid` and `Pins` are writable; and `SetFlags(RF_Transactional)` permanently mutates flags.

The capability is worth having — I would want it scoped to `UEdGraphNode` with an allowlist, `reconstruct` opt-in, and no save. If you want to take that on I am happy to review the shape before you write it; otherwise it stays on my list.

Confirmed your claim that no `/Private/` engine headers are used, incidentally — checked both files' include lists.

Leaving open. Credit in the notes when part (a) ships.

tumourlove:
Part 1 shipped in [v0.22.0](https://github.com/tumourlove/monolith/releases/tag/v0.22.0). I land contributor fixes as my own commits rather than merging the branch — I keep the shipped history single-author for release integrity, and credit you in the release notes instead. No reflection on the patch.

The ICH write tier is in. `set_component_property` now resolves components declared in a **parent Blueprint's** construction script by walking the chain for the defining `USCS_Node`, building an `FComponentKey`, and get-or-creating an Inheritable Component Handler override — with the structural-modify and recompile handshake, without which the override silently does not persist. Your approach, essentially unchanged.

It landed as part of something larger. Investigating a separate read bug (#116) turned up **five** independent component resolvers in this module, three of which disagreed about which CDO to read — your write path had it right and the read path did not. So rather than adding a third tier that only the write path knew about, there is now one resolver used by both, reporting `source` as `scs` | `cdo_native` | `ich_override` | `inherited_scs` | `parent_cdo_fallback`.

Worth knowing if you build on this: **the old `ich` source could never fire.** `get_inherited_component_override` compared Inheritable Component Handler records against a CDO component's name, which are different objects. It is `ich_override` now and it actually reports.

**Part 2 — `set_graph_node_property` / `get_graph_node_properties` — is deferred, not rejected.** The need is real: there is no general way to read or write a graph node's own UPROPERTYs, and calling `ReconstructNode()` after such a write is the genuinely new insight in the submission. What I am not comfortable shipping as submitted is the shape — it accepts an arbitrary object path and writes an arbitrary reflected property, with `save` defaulting to true, which is set-any-property-on-any-object with no allowlist and no scoping. Four typed equivalents already ship for specific node families, and the general version wants a narrower contract first. It is logged with your `ReconstructNode()` observation recorded, so nothing is lost.

One incidental thing you may want for future PRs: the diff reads as +1224/-843 but is really about +383. The repo stores LF and the branch came in with CRLF, so git scored both `.cpp` files as whole-file rewrites. `MonolithBlueprintComponentActions.cpp` is genuinely +49/-2.

Thomasbehan:
Following proper git standards, you should have done a code review here inline, I fix to your repo standards and you merge in... Release integrity is a shitty excuse to be the only contributor as you can easily do releases by tags on main which you control since noone else can tag main.
Latest != Latest release.

Any further contributions I'll hold back on and fork this repo instead without this bad contribution practices, I'll happily return if your contribution guidelines change 🫡

PR #103, opened 2026-07-24
Added compatibility with Unreal Engine 5.6

Puche:
This PR extends plugin support (previously 5.7–5.8 only) to also compile and run on UE 5.6, using preprocessor guards (ENGINE_MAJOR_VERSION / ENGINE_MINOR_VERSION) in the modules where the engine API changed between versions:

Animation / Control Rig: alternate include paths for ControlRigBlueprintLegacy.h (not a separate header in 5.6), accessing RigVMClient without the IRigVMAssetInterface indirection, and a manual fallback to get the entry node's output pin in state machines (GetOutputPin() wasn't available in 5.6).
Pose Search: the biggest chunk of the change. In 5.6, the Pose Search database doesn't use the unified FPoseSearchDatabaseAnimationAsset struct — instead it stores per-type concrete structs (FPoseSearchDatabaseSequence, BlendSpace, AnimComposite, AnimMontage, MultiAnimAsset) wrapped in an FInstancedStruct. Added a helper function (MakePoseSearchDatabaseEntry) that builds the right entry based on asset type, and updated all database read/write call sites (get, add, remove, stats, tags, validation) to work with both paths.
AI Indexer: small unrelated fix — corrects a FJsonSerializer::Serialize call that needed Props.ToSharedRef() instead of passing the TSharedPtr directly.

Overall: 26 files changed, +362/-22 lines, all additive via #if/#else with no behavior change on 5.7/5.8.


tumourlove:
Thanks for this, and sorry it sat unanswered for so long — that was on me, not on the work.

I am declining it, and I want to give you the real reason rather than a vague one.

**I cannot verify it.** UE 5.6 is not installed here, there is no CI, and the dual-engine release pipeline builds 5.7 and 5.8 only. Merging this would mean shipping ~360 lines of `#if`-guarded code that has never been compiled, in a plugin whose release process otherwise gates on two full builds, a full-unity collision pass, and a DLL-import smoke test per engine. I would be asserting support I have no way to stand behind, and the first 5.6 user to hit a broken guard would be finding it for me.

**The PR also shows signs of not having been compile-verified on 5.6 itself** — there are guarded paths that would not build as written. That is not a criticism of the effort; it is the same problem I have, from the other side. It is genuinely hard to get right without the engine in front of you.

**And "compiles on 5.6" is less than it sounds.** The release pipeline would still ship no 5.6 zip, so downstream 5.6 users would have to build from source anyway, while every future change in this plugin would have to be reasoned about on three engines instead of two. That is a standing tax on every subsequent PR, paid for a configuration nobody can test.

Two specific notes, because your analysis deserves them:

- **The pose-search work is correct.** 5.6 really does store per-type concrete structs rather than the unified `FPoseSearchDatabaseAnimationAsset`, and your `MakePoseSearchDatabaseEntry` helper is the right shape for bridging that. That was the largest and most careful part of the submission.
- **The `Props.ToSharedRef()` change is not a regression** — I want to correct that on the record, because an earlier internal note of mine said it was, and that note was wrong. Both forms are valid `FJsonSerializer::Serialize` overloads and master already uses `.ToSharedRef()` at around thirty call sites. On a null pointer it is if anything marginally better: a clean `check` rather than an access violation. I am not taking it only because it is a no-op on the engines I ship.

If 5.6 support becomes something several people need, the blocker is testability rather than the patch — a 5.6 host project plus a way to build it in the release matrix. I would revisit it then, and this PR would be the starting point.

Closing, with thanks.

PR #109, opened 2026-07-24
niagara: optional link topology in get_module_graph

whalemenace:
`get_module_graph` currently serializes nodes and pins with a `linked_count` per pin but no edge information, so consumers that need actual graph topology (branch tracing through static switches, dataflow analysis, dead-branch detection) have to fall back to exporting the asset to T3D and parsing `LinkedTo` out of that.

This adds an opt-in `links` boolean param (default `false`, output unchanged when absent). When `true`, each pin gains its `pin_id` (the `UEdGraphPin::PinId` GUID, since pin names are not unique on a node) and a `linked_to` array of `{node_guid, pin_id, pin_name}` for each entry in `UEdGraphPin::LinkedTo`. Null-safe via `GetOwningNodeUnchecked()`.

Tested on UE 5.8 against Niagara module scripts with static-switch topology; verified the emitted edges match the asset's T3D export exactly (396/396 edges on a 264-node / 1,027-pin production module, symmetric when keyed by `(node_guid, pin_id)`).

One field note for consumers worth knowing: `PinId` alone is NOT unique within a real graph (copy-paste history duplicates them — the test module had 1,027 pins with only 869 unique ids), which is why `linked_to` entries carry `node_guid` and edges should be keyed by the `(node_guid, pin_id)` pair.

### DIFF
--- a/Source/MonolithNiagara/Private/MonolithNiagaraActions.cpp
+++ b/Source/MonolithNiagara/Private/MonolithNiagaraActions.cpp
@@ -2615,6 +2615,7 @@ void FMonolithNiagaraActions::RegisterActions(FMonolithToolRegistry& Registry)
 		FMonolithActionHandler::CreateStatic(&HandleGetModuleGraph),
 		FParamSchemaBuilder()
 			.RequiredAssetPath(TEXT("script_path"), TEXT("Module script asset path"))
+			.Optional(TEXT("links"), TEXT("boolean"), TEXT("Include per-pin pin_id and linked_to [{node_guid, pin_id, pin_name}] arrays for link topology (default false)"))
 			.Build());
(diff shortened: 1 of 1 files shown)


tumourlove:
Shipped in v0.21.3 — thanks. The gap was real: `get_module_graph` emitted `linked_count` and nothing else, so anyone needing actual topology had to export T3D and parse `LinkedTo` out of it.

One change from your version: instead of a per-pin `linked_to` array I emit a single top-level `edges` array, deduplicated on the endpoint pair. The pin walk sees every edge twice, once from each end, so the per-pin shape doubled the payload on graphs that are already large — `get_module_graph` has no pagination yet, which I have noted as a separate follow-up.

Your field note about `PinId` not being unique within a graph was the useful part and I kept it — edges are keyed by the `(node_guid, pin_id)` pair, and that is called out in the docs so consumers do not get bitten by copy-paste duplicates. The opt-in default is preserved: output is byte-identical when `links` is absent.

I land contributor fixes as my own commits rather than merging the branch — shipped history stays single-author, and the reporter gets credited in the release notes.

PR #110, opened 2026-07-24
UI: accept string-encoded values for Margin and Vector4 properties

Thomasbehan:
## Summary

`ui::set_widget_property` cannot set any **Margin** or **Vector4** property. The value is always rejected with

```
ParseFailed on Border.Padding (expected Margin: scalar, [l,t,r,b], or {left,top,right,bottom})
```

…even though the caller passed exactly one of the listed shapes.

## Cause

The `value` param is declared `string` in the action schema, so a caller writing JSON (`[22,14,22,14]`, `{"left":26,...}`) or a bare number (`20`) delivers it as **text**, and the JSON value reaching the parser is `EJson::String`.

In `UIReflectionHelper.cpp` the sibling parsers already anticipate that — `ParseVector2D` accepts `"x,y"` and `ParseLinearColor` accepts `"#RRGGBBAA"` — but `ParseMargin` and `ParseVector4` only branch on `Number` / `Array` / `Object`, so they fall through to `return false` for every string. The advertised shapes are therefore unreachable through `set_widget_property`, and the error message lists shapes that cannot get through, which sends callers hunting for a value-format problem that does not exist.

The registry's K4 string-encoded-param recovery does not help here: it only coerces params the **schema** declares as `array`/`object`, and this one is declared `string`.

## Fix

- Add a `CoerceStringToJson` helper next to the other struct parsers: if the value is a string that looks like JSON (`{...}` / `[...]`) it is re-parsed; a bare numeric string becomes a JSON number (the documented "scalar" form for `FMargin`). Anything else is returned untouched.
- Apply it at the top of `ParseMargin` and `ParseVector4`.
- Also accept the comma-separated text form (`"l,t,r,b"` / `"x,y,z,w"`), matching `ParseVector2D`'s existing behaviour.

`ParseSlateColor` needed no change — it delegates to `ParseLinearColor`, which is already string-tolerant.

Behaviour for callers already sending real JSON types is unchanged: non-string values skip the helper entirely, and a string that is neither JSON nor numeric is passed through so the existing failure path still reports it.

## Testing

Built against **UE 5.7** (`WWJEditor Win64 Development`, clean compile) and exercised live over the MCP server:

| Call | Before | After |
|---|---|---|
| `set_widget_property` `Border.Padding` = `{"left":26,"top":10,"right":26,"bottom":14}` | ParseFailed | applied |
| `set_widget_property` `Border.Padding` = `[22,14,22,14]` | ParseFailed | applied |
| `set_widget_property` `Border.Padding` = `20` (uniform scalar) | ParseFailed | applied |
| `ColorAndOpacity`, `Percent`, `Font`, `Justification` (regression check) | worked | unchanged |

### DIFF
--- a/Source/MonolithUI/Private/Registry/UIReflectionHelper.cpp
+++ b/Source/MonolithUI/Private/Registry/UIReflectionHelper.cpp
@@ -12,6 +12,7 @@
 #include "Registry/UIReflectionHelper.h"
 
 #include "Dom/JsonObject.h"
+#include "Dom/JsonValue.h"
 #include "Layout/Margin.h"
 #include "Math/Color.h"
 #include "Math/Vector2D.h"
@@ -21,6 +22,8 @@
 #include "Registry/UIPropertyAllowlist.h"
 #include "Registry/UIPropertyPathCache.h"
 #include "Registry/UITypeRegistry.h"
+#include "Serialization/JsonReader.h"
+#include "Serialization/JsonSerializer.h"
 #include "Styling/SlateColor.h"
 #include "UObject/Class.h"
 #include "UObject/EnumProperty.h"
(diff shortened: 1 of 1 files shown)


tumourlove:
Shipped in v0.21.3 — thanks, your diagnosis was exactly right. `value` is declared `string` in the action schema, so `ParseMargin`/`ParseVector4` only ever saw `EJson::String` and fell straight through to failure, while the error text helpfully listed the three shapes that could not get through. Pointing at the `ParseVector2D`/`ParseLinearColor` asymmetry is what made it obvious.

I went a bit further than the patch: `value` is now declared `any`, which is the actual root cause, and I kept string coercion on top of it — the proxy and CLI serialise nested params to text regardless, so both are load-bearing. Comma-separated forms are in too, and I tightened the numeric check while I was there (`FString::IsNumeric` accepts a bare `"-"` and coerces it to 0.0).

For future PRs: the branch flipped the whole file LF to CRLF, which turned an ~80-line change into a 1,100-line diff and made it hard to review. I have added a `.gitattributes` so it will not happen again.

I land contributor fixes as my own commits rather than merging the branch — I keep the shipped history single-author for release integrity, and credit the reporter in the release notes instead. No reflection on the patch; yours was solid.

Thomasbehan:
Following proper git standards, you should have done a code review here inline, I fix to your repo standards and you merge in... Release integrity is a shitty excuse to be the only contributor as you can easily do releases by tags on main which you control since noone else can tag main.
Latest != Latest release.

Any further contributions I'll hold back on and fork this repo instead without this bad contribution practices, I'll happily return if your contribution guidelines change 🫡

Issue #117, opened 2026-07-27
If Unreal Editor crashes or is terminated while Monolith is performing the initial/full project index, Monolith starts the entire index again from zero on the next launch.


Hvizeu:
## Environment

- Monolith: v0.21.3
- Unreal Engine: 5.8
- Platform: Windows
- Large Unreal project with many assets

## Steps to reproduce

1. Open a project that does not yet have a completed Monolith index.
2. Allow the full project index to process several batches.
3. Close or crash Unreal Editor before indexing finishes.
4. Reopen the project.
5. Observe that Monolith begins parsing every asset again from the start.

## Actual behavior

Monolith discards all previously indexed progress and performs a completely new index.

The likely cause is the full-index startup flow:

- `StartFullIndex()` unconditionally calls `ResetDatabase()`.
- `ResetDatabase()` drops the index tables, including already committed batches.
- `last_full_index` is only written after the entire indexing process completes.
- After a crash, that marker is absent, so `ShouldAutoIndex()` starts another full index.
- Starting that index resets the database and destroys the recoverable progress.

The editor crash itself does not need to be caused by Monolith. Any interruption during indexing produces the same loss of progress.

## Expected behavior

Full indexing should be restart-safe:

- Persist an `in_progress` full-index state.
- Commit a durable checkpoint after every successfully committed batch or asset.
- On restart, retain the existing database.
- Rescan lightweight asset metadata and resume only pending, failed, or changed assets.
- Rerun post-processing passes safely when necessary.
- Clear the in-progress state and write `last_full_index` only after successful completion.
- Preserve the existing explicit **Force Reindex** behavior so users can intentionally wipe and rebuild the index.

Checkpoint updates should be committed atomically with the corresponding indexed data so the database cannot report an asset as completed when its data was rolled back.

## Impact

On large projects, a single Unreal Editor, driver, or machine crash can waste hours of indexing work. Repeated crashes can prevent Monolith from ever producing a usable project index.

tumourlove:
Can you send me a crash log please, just to rule out a single asset causing the crash xo

Hvizeu:
Unfortunately, I can’t share the crash log because it contains project information covered by an NDA. Scrubbing it thoroughly enough to make it legally shareable would also take a significant amount of time.

[...]

PR #120, opened 2026-07-30
fix(index): resume interrupted full indexing

Hvizeu:
## What and why

Full-project indexing currently starts over after an editor shutdown or crash because an interrupted run never writes `last_full_index`. On large projects, repeated restarts can prevent the initial index from ever completing.

This change adds a schema-v3 recovery lifecycle that:

- stores per-asset progress by package path, saved hash, and indexer name
- commits indexed data and its progress checkpoint in the same SQLite transaction
- refreshes metadata after a restart, removes deleted packages, and queues only unfinished or hash-invalidated work
- tracks post-pass completion and reruns invalidated passes when indexed assets change
- processes animation assets individually on compiler-idle editor ticks so interrupted animation indexing can resume
- requires all registered full-index post-passes to complete before publishing `last_full_index`
- leaves failed work pending and rolls back incomplete transactions
- preserves the existing forced-reindex behavior, which intentionally clears data and recovery state

## Checklist

- [x] Builds on **UE 5.7** (the compile floor)
- [x] Builds on **UE 5.8**
- [x] Any engine API newer than 5.7 sits behind an `ENGINE_MINOR_VERSION` gate, with a working 5.7 path (N/A: no newer engine API added)
- [x] No unrelated whitespace, line-ending, or reformatting churn in the diff
- [x] Specs updated (`Docs/specs/SPEC_MonolithIndex.md`)
- [x] `CHANGELOG.md` updated under `## [Unreleased]`
- [x] Recovery behavior verified with editor automation on both supported engine versions
- [x] New or changed actions verified in-editor (N/A: no actions changed)

## Testing notes

- `git diff --check upstream/master` passes.
- `Monolith.Index.Recovery.CheckpointTransaction`, `Monolith.Index.Recovery.Lifecycle`, and `Monolith.Index.Recovery.PostPassCleanup` pass under **UE 5.7**: **3 succeeded, 0 failed**.
- The same three tests pass under **UE 5.8**: **3 succeeded, 0 failed**.
- **UE 5.7:** `Editor Win64 Development` completed the final 49-action build successfully, including the changed Monolith sources, tests, dependent plugin modules, and DLL links.
- **UE 5.8:** `Editor Win64 Development` completed the final 23-action build successfully, including the changed Monolith sources, tests, dependent plugin modules, and DLL links.
- The remaining compiler warnings are pre-existing deprecations or warnings outside the changed code.


tumourlove:
Shipped in [v0.22.0](https://github.com/tumourlove/monolith/releases/tag/v0.22.0). I land contributor fixes as my own commits rather than merging the branch — I keep the shipped history single-author for release integrity, and credit you in the release notes instead. No reflection on the patch.

Your diagnosis was exactly right and I verified every link: `StartFullIndex()` unconditionally resetting, `last_full_index` written only at the end, `ShouldAutoIndex()` therefore restarting and wiping the recoverable progress. The schema-v3 lifecycle and the atomic-checkpoint idea are yours and are preserved.

I did land it smaller, and there were three things I changed rather than took:

**The poison pill.** This is the important one. Your version records completions but never *attempts*, so a hard crash leaves no trace and the resume re-queues the asset that killed the editor — a crash-every-launch loop, which is worse than the bug. It is made sharper by the `bDeferFirstTimeIndex` bypass on the resume path: that flag is the documented escape hatch for exactly this crash class, and overriding it in the crash case leaves deleting `ProjectIndex.db` by hand as the only recovery. The shipped version counts attempts, skips after repeated failures, and surfaces the skipped asset through `project get_stats`.

Subtlety worth recording, because I got it wrong first and a review caught it: the attempt marker has to commit in its **own** transaction *before* the batch work. Written inside the batch transaction it is not durable — SQLite runs a DELETE journal here, so the crash rolls the marker back along with the work and the counter still reads zero on resume. The pill can never fire in the one case it exists for.

**The completion gate.** `Errors > 0 || DeepErrors > 0` blocking `last_full_index` reproduces #117's symptom from the other direction. `DeepErrors++` fires routinely — assets referencing classes from disabled plugins, redirector stubs — and the codebase already carries an SEH guard specifically because some animation assets crash on load. One such asset and the marker is never written, so every launch runs a full index. Only structural failures fail the gate now.

**The animation indexer.** Per-asset transactions and `BatchSize = 1` are a throughput regression bundled into a correctness fix, so I kept the existing dispatch model.

Also not taken: post-pass resume checkpoints (they self-invalidate on the common interruption case), the `ClearFullIndexPostPassData` name→SQL map (wrong layer, and wrong for `AI`), and activating the three dead sentinel indexers — that last one is a real latent bug you surfaced, but it changes the runtime and memory profile of every full index, so it gets its own change and its own verification rather than riding along. It is logged.

One note on the checklist: the three automation tests exercise `FMonolithIndexDatabase` in isolation — none of them drives an interrupted index, a resume, or a crash. The shipped tests include a rollback-durability assertion, because "the counter increments" cannot distinguish the durable implementation from the broken one.

PR #124, opened 2026-08-03
fix(editor): prevent PIE compile-error modal deadlocks

Hvizeu:
## Problem

When editor.start_pie encounters loaded Blueprints with unresolved compiler errors, Unreal can open a blocking confirmation dialog. Monolith runs in-process, so that modal blocks the game thread and makes the current request and later MCP calls appear deadlocked.

## Fix

- Add on_compile_errors to editor.start_pie with two explicit policies.
- refuse is the default: return a structured error containing the offending Blueprint names and paths without starting PIE.
- suppress starts PIE under a narrowly scoped unattended-script guard so Unreal cannot open the blocking confirmation modal.
- Return the selected policy, error count, and Blueprint list on successful starts.
- Reject unknown policies as invalid parameters.
- Document the new request and response contract.
- Add focused automation for invalid policies and structured refusal.

## Checklist

- [x] UE 5.7 editor build
- [x] UE 5.8 editor build
- [x] Focused editor automation
- [x] Live refuse and suppress verification
- [x] API reference, module spec, and changelog updated
- [x] Project-agnostic implementation and public description

## Validation

- Monolith.Editor.PIE automation passed 2/2.
- Refuse mode returned immediately with structured error data and left MCP responsive.
- Suppress mode started PIE successfully while MCP remained responsive.
- The UE 5.7 and UE 5.8 UnrealEditor targets compiled successfully.
- Full plugin packaging reaches and compiles the changed editor target; a pre-existing MonolithAudioRuntime failure outside this diff still blocks the later packaged-game target.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **`editor start_pie` no longer strands MCP behind the Blueprint compile-error dialog.** The action now pre-flights loaded Blueprints using the engine's own unresolved-error condition. Its default `on_compile_errors: "refuse"` policy returns the offending `{name, path}` entries without starting PIE; `"suppress"` explicitly starts anyway under a scoped unattended-script guard, so the confirmation modal cannot block the game thread and the in-process MCP server.
+
 ## [0.22.0] - 2026-08-01
 
 ### Internal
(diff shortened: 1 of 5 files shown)


tumourlove:
Shipped in v0.23.0.

Reimplemented at HEAD rather than merged, per the usual single-author history. Both policies are as you specified, with `refuse` the default so nothing changes for existing callers who do not ask for it.

One deviation worth naming: rather than approximating the condition, the pre-flight now tests the same thing the engine does -- `BS_Error && bDisplayCompilePIEWarning`, matching `FInternalPlayLevelUtils::ResolveDirtyBlueprints` -- so the refusal set is exactly the set that would have raised the dialog, no wider. The suppress path keeps your narrow scoping and is RAII, so it cannot leak past an early return.

You were right that the failure mode is the interesting part: in-process means that modal starves the MCP server, so it does not present as a dialog, it presents as a hang.

PR #126, opened 2026-08-04
fix(audio): make runtime module self-contained for packaged targets

kunkunGames:
## Summary

Make `MonolithAudioRuntime` self-contained for packaged targets by including the public headers that own every engine symbol used by its implementation files.

## Problem

The runtime module compiled in Editor builds because declarations arrived transitively through shared-PCH or unity-build context. Its implementation files directly use `GEngine`, `UEngine::GetWorldFromContextObject`, `EGetWorldErrorMode`, and `IMPLEMENT_MODULE`, but did not include the headers that declare those symbols.

That implicit dependency is unsafe for monolithic `UnrealGame` targets and can break packaged Development or Shipping builds even when the Editor target succeeds.

## Solution

- include `Engine/Engine.h` in the runtime perception implementation
- include `Modules/ModuleManager.h` in the runtime module implementation
- document the include-ownership and packaged-target contract

This is deliberately a translation-unit ownership fix. It adds no module dependency and changes no runtime control flow, asset behavior, action surface, or configuration.

## Verification

- UE 5.7 `RunUAT BuildPlugin -TargetPlatforms=Win64 -Rocket` passed Editor Development, UnrealGame Development, and UnrealGame Shipping; UAT exit code 0
- UE 5.8 `RunUAT BuildPlugin -TargetPlatforms=Win64 -Rocket` passed Editor Development, UnrealGame Development, and UnrealGame Shipping; UAT exit code 0
- both packaged-target runs compiled `MonolithAudioPerceptionStatics.cpp` and `MonolithAudioRuntimeModule.cpp` as individual non-unity actions
- both engine roots were resolved from validation-host `.uproject` `EngineAssociation` values
- `git diff --check` passed
- detailed evidence: `Docs/testing/2026-08-04-audio-runtime-packaged-target.md`

## Compatibility and risk

The change is compile-time only and preserves existing runtime behavior. It removes reliance on incidental Editor include order, so the affected packaged build paths become more deterministic without requiring migration.

## Visual evidence

Not applicable: this is a compile dependency correction with no visual or editor-facing UI change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **`MonolithAudioRuntime` now compiles as a self-contained packaged-target module.** Its translation units include the owning public headers for `IMPLEMENT_MODULE`, `GEngine`, `UEngine::GetWorldFromContextObject`, and `EGetWorldErrorMode` instead of relying on Editor shared-PCH or unity-build transitive includes.
+
 ## [0.22.0] - 2026-08-01
 
 ### Internal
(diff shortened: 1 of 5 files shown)


tumourlove:
Shipped in v0.23.0, though not from this branch -- #136 by @itismyfield added the same two includes (`Engine/Engine.h` and `Modules/ModuleManager.h`) as part of a 106-file sweep, and that landed first. Your diagnosis of the packaged-target risk was right, and it is what made the sweep worth taking whole.

Closing as fixed. Credited in the release notes alongside your other four.

Issue #121, opened 2026-08-01
Full support for 5.8.1


aggitti:
Upon fresh compilation of the plugin in an Unreal Engine 5.8.1 version, the following Warnings are shown, as the plugin compiles upon the 1st project run/start.

```
\Source\MonolithIndex\Private\Indexers\GenericAssetIndexer.cpp(82,70): warning C4996: 'UMaterialExpressionTextureBase::GetSamplerTypeForTexture': Use MaterialExpressionUtils::GetSamplerTypeForTexture instead. - Please update your code to the new API before upgrading to the next release, otherwise your project will no longer compile.

\Source\MonolithIndex\Private\Actions\ProjectExportAssetTextAction.cpp(81,3): warning C4996: 'GetObjectsWithOuter': GetObjectsWithOuter with a boolean bIncludeNestedObjects has been deprecated - please use the EGetObjectsFlags enum instead - Please update your code to the new API before upgrading to the next release, otherwise your project will no longer compile.

\Source\MonolithBlueprint\Private\MonolithBlueprintActions.cpp(577,6): warning C4996: 'GetObjectsWithOuter': GetObjectsWithOuter with a boolean bIncludeNestedObjects has been deprecated - please use the EGetObjectsFlags enum instead - Please update your code to the new API before upgrading to the next release, otherwise your project will no longer compile.

\Source\MonolithAI\Private\MonolithAIModule.cpp(58,42): warning C4996: 'FCoreDelegates::OnPostEngineInit': OnPostEngineInit has been deprecated. Please use GetOnPostEngineInit() instead. - Please update your code to the new API before upgrading to the next release, otherwise your project will no longer compile.
		PostEngineInitHandle = FCoreDelegates::OnPostEngineInit.AddLambda([this]()

\Source\MonolithAI\Private\MonolithAIModule.cpp(85,19): warning C4996: 'FCoreDelegates::OnPostEngineInit': OnPostEngineInit has been deprecated. Please use GetOnPostEngineInit() instead. - Please update your code to the new API before upgrading to the next release, otherwise your project will no longer compile.
		FCoreDelegates::OnPostEngineInit.Remove(PostEngineInitHandle);

\Source\MonolithAnimation\Private\MonolithMirrorTableActions.cpp(148,7): warning C4996: 'UMirrorDataTable::FindReplaceMirroredNames': Please use UpdateFromFindReplaceExpressions() instead. - Please update your code to the new API before upgrading to the next release, otherwise your project will no longer compile.
[...]

PR #127, opened 2026-08-04
fix(ue58): use versioned material sampler API

kunkunGames:
## Summary

Add one reusable UE 5.7/5.8 compatibility boundary for material sampler inference and route both Monolith consumers through it.

## Problem

UE 5.8 moved `GetSamplerTypeForTexture` from `UMaterialExpressionTextureBase` to `MaterialExpressionUtils` and deprecated the former entry point. UE 5.7 does not expose `MaterialExpressionUtils.h`, so an unconditional migration would break Monolith's supported compile floor.

The deprecated call existed independently in `MonolithIndex` and `MonolithMaterial`, which duplicated the compatibility problem and produced both sampler warnings tracked in #121.

## Solution

Introduce `MonolithMaterialSamplerCompat::GetSamplerTypeForTexture` as a small header-only adapter:

- UE 5.7 uses `UMaterialExpressionTextureBase::GetSamplerTypeForTexture`
- UE 5.8 and newer use `MaterialExpressionUtils::GetSamplerTypeForTexture`
- both existing consumers call the same Monolith compatibility API

Centralizing the version gate avoids repeated preprocessor branches and gives future sampler consumers one stable extension point. Returned metadata and runtime behavior remain unchanged.

## Verification

- UE 5.7 `RunUAT BuildPlugin -NoTargetPlatforms -Rocket`: 434/434 Editor actions passed; UAT exit code 0
- UE 5.8 `RunUAT BuildPlugin -NoTargetPlatforms -Rocket`: 434/434 Editor actions passed; UAT exit code 0
- both builds compiled `GenericAssetIndexer.cpp` and `MonolithMaterialActions.cpp` as individual actions
- the UE 5.8 build log contains zero `GetSamplerTypeForTexture` C4996 warnings
- `git diff --check` passed
- detailed evidence: `Docs/testing/2026-08-04-ue58-material-sampler-api.md`

## Compatibility and risk

This PR addresses only the two material-sampler deprecations in #121; unrelated UE 5.8 warnings remain out of scope. The adapter preserves UE 5.7 support and does not change public MCP actions, serialized data, or configuration.

## Visual evidence

Not applicable: this is a compile-time engine API compatibility change with no visual or editor-facing UI behavior.


tumourlove:
Shipped in v0.23.0.

Reimplemented at HEAD as `MonolithMaterialSamplerCompat` rather than merging the branch -- I keep the shipped history single-author. The design is yours: one header-only adapter, both consumers routed through it, so the version gate exists in one place.

Verified `MaterialExpressionUtils::GetSamplerTypeForTexture` against the installed 5.8 headers (`Runtime/Engine/Public/Materials/MaterialExpressionUtils.h:25`, `ENGINE_API`) before taking it. A full 5.8 build now reports zero C4996 warnings, so this plus the rest of the #121 sweep closes that out completely.

PR #128, opened 2026-08-04
fix(proxy): fail when native proxy publication fails

kunkunGames:
## Summary

Make the Windows native-proxy build report success only after a complete executable has been safely published to its final destination.

## Problem

Both proxy build entry points previously compiled in the source tree, copied directly over `Binaries\\monolith_proxy.exe`, ignored publication failures, and printed success unconditionally. A locked or unwritable destination could therefore leave the previous executable in place while returning a misleading successful result. Direct replacement also exposed the live binary to partial-publication risk.

## Solution

- keep `Tools\\MonolithProxy\\build_proxy.bat` as the single authoritative implementation
- make `Tools\\MonolithProxy\\build.bat` a compatibility delegate that preserves the authoritative exit code
- compile in an invocation-owned private staging directory
- copy the completed executable to a unique candidate beside the destination and verify its byte count
- make the final same-directory move the only replacement operation
- validate every toolchain, directory, compile, copy, move, existence, and size gate before printing success
- on failure, preserve the existing proxy and clean only invocation-owned staging/candidate paths

Explicit source, output, and staging overrides let the regression harness exercise the production batch workflow without maintaining a second test-only implementation.

## Verification

- Windows PowerShell `5.1.26100.8655`: `Scripts\\test_proxy_build.ps1` passed
- PowerShell `7.5.5`: `Scripts\\test_proxy_build.ps1` passed
- both public batch entry points built and published a non-empty proxy
- injected C++ compilation failure returned non-zero, printed no success message, and preserved the existing sentinel executable byte-for-byte
- a locked destination forced native replacement failure; the command returned non-zero, printed no success message, preserved the sentinel byte-for-byte, and removed candidate/staging artifacts
- `git diff --check` passed
- detailed evidence: `Docs/testing/2026-08-04-native-proxy-publication-failure.md`

## Compatibility and risk

This changes only the Windows native-proxy build and publication workflow. MCP protocol behavior, Unreal modules, assets, and runtime proxy logic are unchanged. Existing callers can continue using either batch entry point, but failed publication now correctly produces a non-zero exit code.

## Visual evidence

Not applicable: this is command-line build/publication reliability work with no visual or editor-facing UI change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **Native proxy publication now fails closed.** Both Windows build entry points share one implementation that compiles in private staging, validates a destination-side candidate, and checks final replacement before reporting success. Compile or publication failures return non-zero, preserve an existing proxy before replacement, and clean only invocation-owned staging and candidate files.
+
 ## [0.22.0] - 2026-08-01
 
 ### Internal
--- a/Docs/SPEC_CORE.md
+++ b/Docs/SPEC_CORE.md
@@ -453,6 +453,12 @@ Everything lives in one place: `YourProject/Plugins/Monolith/`
 
 This folder is both the working copy and the git repo (`git@github.com:tumourlove/monolith.git`). Edit, build, commit, push, and release all happen here — no file copying.
 
+#### Native proxy publication
+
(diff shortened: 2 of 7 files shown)


tumourlove:
Shipped in v0.23.0.

Reimplemented at HEAD rather than merged, per the usual single-author history. Kept your structure intact: `build_proxy.bat` as the one authoritative implementation, `build.bat` as a delegate preserving the exit code, staged compile, size-checked candidate, same-directory move as the only replacement.

Ran your harness on Windows PowerShell 5.1 -- both entry points publish, and the injected compile failure and the locked destination both preserve the previous proxy byte-for-byte.

Worth noting this is the second build script in this repo caught printing success over a failure; the `monolith_query` one had the same shape last release. Yours is the better fix, because it also removes the partial-publication window rather than just reporting it.

PR #131, opened 2026-08-04
feat(localization): add StringTable discovery and validation

kunkunGames:
## Summary

Add a cohesive, read-only localization preflight surface for culture discovery and bounded StringTable discovery, readback, and validation without mutating or dirtying packages.

## Problem

Monolith already exposes StringTable authoring through the broad `blueprint` namespace, but automation has no focused read-only contract for finding tables, paging large entry sets, bounding metadata/text payloads, or proving that validation covered the whole asset. Callers currently have to load and reconstruct this state themselves, and a truncated scan can be mistaken for a valid table.

## Solution

Add four actions under a dedicated read-only and idempotent `localization` dispatcher:

| Action | Purpose |
| --- | --- |
| `list_cultures` | Stable bounded Unreal culture discovery, including explicit-root resolution and unresolved names |
| `list_string_tables` | Stable Asset Registry discovery with bounded pagination and optional returned-page details |
| `get_string_table` | Exclusive-key entry pagination with independent metadata-row and text-length budgets |
| `validate_string_table` | Deterministic bounded key/source issues, paginated diagnostics, and explicit completeness |

All asset inputs require canonical mounted package paths or matching top-level object paths. Culture lists, asset pages, entry pages, metadata rows, text values, validation scans, and issue pages have documented hard bounds. Readback separates `has_more_entries`, `all_entries_covered`, and `metadata_complete`; validation reports `valid=true` only when the scan is complete and has zero errors.

The implementation keeps memory bounded while finding the lexicographically smallest entry keys, rather than materializing an unbounded sorted copy. It registers before the config-authoring gate because culture and StringTable inspection do not depend on config mutation, so these actions remain available when `bEnableConfig=false`.

The existing `blueprint` StringTable actions remain the authoring surface. This PR intentionally adds no create, set, remove, import, export, transaction, dirty, or save action.

## Verification

- UE 5.7 `RunUAT BuildPlugin -NoTargetPlatforms -Rocket`: 436/436 build actions passed; UAT exit code 0
- UE 5.8 `RunUAT BuildPlugin -NoTargetPlatforms -Rocket`: 436/436 build actions passed; UAT exit code 0
- final UE 5.7 Editor target rebuild: 436/436 build actions passed; UBT exit code 0
- final UE 5.8 Editor target rebuild: 436/436 build actions passed; UBT exit code 0
- UE 5.7 `Automation RunTests Monolith.Localization.Read`: 3/3 passed, zero failures/skips, exit code 0
- UE 5.8 `Automation RunTests Monolith.Localization.Read`: 3/3 passed, zero failures/skips, exit code 0
- implementation scan found exactly four localization actions and no production transaction, save, mutation, package-dirty, mutable StringTable, or asset create/delete calls
- new source and tests contain no anonymous namespaces, `using namespace` directives, or individual `using` declarations
- `git diff --check` passed
- detailed evidence: `Docs/testing/2026-08-04-string-table-discovery-validation.md`

The first focused run in each engine also disproved a proposed case-insensitive duplicate-key audit: Unreal's `FTextKey` identity does not retain simultaneous `Case` and `case` rows in a StringTable. The unreachable check and misleading test were removed rather than reporting a diagnostic from data the engine cannot preserve; the final runs above are the post-correction 3/3 gates.

## Compatibility and risk

This PR adds a direct `AssetRegistry` dependency to the existing editor module `MonolithConfig`; it does not add a runtime module or change the existing `blueprint` StringTable authoring contract. It also reconciles the stale documented `config` count with the seven actions already registered by source before adding the four-action localization namespace.

Large tables remain bounded by hard caps and explicit continuation/completeness fields. Invalid paths, wrong asset types, load failures, and incomplete validation fail explicitly; no fallback asset, widened path, substitute data, or silent legacy branch is used.

## Visual evidence

Not applicable: this PR adds headless action handlers, schemas, automation tests, documentation, and routing guidance with no visual or editor-facing UI change.


tumourlove:
Shipped in v0.23.0 -- all four actions on a new read-only `localization` namespace.

Reimplemented at HEAD rather than merged, per the usual single-author history. Bounds and completeness semantics kept as designed, including `validate_string_table` reporting `valid=true` only when the scan actually covered every entry, and treating a truncated scan as an error in its own right. Registered ahead of the `bEnableConfig` gate, as you had it.

I also kept your finding out of the implementation: the case-insensitive duplicate-key audit is genuinely unreachable, because `FTextKey` identity will not retain both `Case` and `case`. Removing an unreachable check and the test that pretended to cover it was the right call, and I did not re-add either. 3/3 green on 5.7.

PR #133, opened 2026-08-10
fix(editor): create_empty_map reports current_world and gains an opt-in open param

whalemenace:
`create_empty_map` saves a new UWorld but leaves the editor on whatever world was open, and nothing in the response says so. Agents chain create→populate and silently edit the wrong world — we've shipped actors into `/Temp/Untitled_N` this way, and on 0.22.0 the failure mode is quieter and worse: you edit whatever map happened to be open.

Changes:
- Response gains `current_world` (the editor world after the call) and `opened`, and the success message states explicitly that the open world is unchanged and how to open the new one.
- Opt-in `open` param (default `false`, no behavior change) loads the map after creation via `ULevelEditorSubsystem::LoadLevel` — same semantics as `editor::load_level`, one call instead of the create/load pair every agent writes anyway.

Verified live on 0.22.0: without `open`, `current_world` correctly reports the untouched world and the message warns; with `open=true`, the editor lands on the new map and a follow-up world query confirms it.

### DIFF
--- a/Source/MonolithEditor/Private/MonolithEditorMapActions.cpp
+++ b/Source/MonolithEditor/Private/MonolithEditorMapActions.cpp
@@ -33,8 +33,10 @@
 #include "AssetToolsModule.h"
 #include "IAssetTools.h"
 
+#include "Editor.h"
 #include "Engine/World.h"
 #include "Factories/WorldFactory.h"
+#include "LevelEditorSubsystem.h"
 
 #include "Misc/PackageName.h"
 #include "Modules/ModuleManager.h"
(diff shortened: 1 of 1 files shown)


tumourlove:
Shipped in v0.23.0.

Reimplemented at HEAD rather than merged, per the usual single-author history. `current_world` and `opened` on every response, the message states plainly that the open world is unchanged, and `open` defaults to `false`.

`open=true` delegates to the existing `load_level` path rather than calling `LoadLevel` directly, so it inherits that action's dirty-map refusal, PIE-teardown guard and stale-resident-world guard instead of quietly having none of them.

"You edit whatever map happened to be open" is exactly the class of bug worth reporting -- it produces no error anywhere, and the damage surfaces much later.

PR #134, opened 2026-08-10
fix(index): editor-exit deadlock — stop-aware waits on every game-thread dispatch

whalemenace:
Repro: fresh install, let the first full index start, QUIT_EDITOR. You get "Indexing was still in progress during shutdown — force-stopped", DDC maintenance completes, then the process hangs forever — window gone, log stalled, kill required.

It's a cross-thread deadlock. The indexing worker parks in unbounded `Event->Wait()`s at fourteen sites, every one waiting on work only the game thread can run (throttle/critical/periodic GC via `AsyncTask(GameThread)`, and every `FMonolithCompilerSafeDispatch` batch/sentinel dispatch, which needs a game-thread *tick*). Meanwhile `Deinitialize()` blocks the game thread in `WaitForCompletion()`. Neither side can ever proceed.

Changes:
- `FIndexingTask::WaitForGameThreadEvent`: all fourteen waits poll in 100ms slices and bail only when stop is requested AND `IsEngineExitRequested()` — a plain user cancel still waits normally, because the game thread is alive and will serve the task. Abandoned pooled events are deliberately leaked (a later `Trigger` on a leaked event is harmless; on a recycled one it's corruption).
- `FMonolithCompilerSafeDispatch` gains an opt-out abort token: an abandoned dispatch's Work payload — whose captures reference the unwound worker stack — is never invoked; the ticker triggers the event and unregisters.
- The registry-scan AsyncTask now owns its state via shared pointer instead of capturing the worker stack, so abandoning it is safe too.
- `Deinitialize` completes the still-Pending `TaskNotification` before destroying it. This assert (`FSlateNotificationManager::ShutdownOnPreExit`: "Missing call to SetComplete?") was always reachable on the force-stop path — the deadlock just hung the editor before it could fire.

Verified on 0.22.0: `monolith_reindex force=true`, quit mid-pass — the "force-stopped" warning fires and the editor exits in under 5 seconds, zero crash artifacts. Before the patch this hung 8+ minutes until killed.

### DIFF
--- a/Source/MonolithIndex/Private/MonolithCompilerSafeDispatch.cpp
+++ b/Source/MonolithIndex/Private/MonolithCompilerSafeDispatch.cpp
@@ -9,7 +9,8 @@
 void FMonolithCompilerSafeDispatch::RunOnGameThreadWhenCompilerIdle(
     TUniqueFunction<void()> Work,
     FEvent* CompletionEvent,
-    float TimeoutSeconds)
+    float TimeoutSeconds,
+    TSharedPtr<TAtomic<bool>>* OutAbortToken)
 {
     // Start time captured by value so each tick can compute elapsed.
     const double StartTime = FPlatformTime::Seconds();
@@ -22,6 +23,7 @@ void FMonolithCompilerSafeDispatch::RunOnGameThreadWhenCompilerIdle(
         FEvent* CompletionEvent = nullptr;
         double StartTime = 0.0;
         float TimeoutSeconds = 120.0f;
+        TSharedPtr<TAtomic<bool>> bAborted;
     };
 
     TSharedPtr<FDispatchState> State = MakeShared<FDispatchState>();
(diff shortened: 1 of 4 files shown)


tumourlove:
Shipped in v0.23.0.

Reimplemented at HEAD rather than merged, per the usual single-author history. All fourteen wait sites go through a stop-aware helper, the abort token is there so an abandoned dispatch's payload never runs against an unwound worker stack, the registry scan owns its state by shared pointer, and `Deinitialize` completes the notification before destroying it.

Your instruction to deliberately leak abandoned pooled events is preserved with a comment explaining why, because it looks exactly like a bug someone will try to tidy up later.

The distinction you drew -- bail only when stop is requested AND the engine is exiting, so a plain user cancel still waits normally -- is what makes this safe, and it is the part a careless version of this fix gets wrong.

PR #135, opened 2026-08-10
feat(editor): capture_pie_movement_clip include_ui — composited game+UI frames

whalemenace:
`capture_pie_movement_clip` frames are scene-only — the viewport `ReadPixels` path never sees UMG/Slate. If you're iterating on UI (which is half of what agents point cameras at), the clip shows an empty scene while the thing you care about is invisible. Our workaround was a hand-rolled python slate-tick ticker spamming `shot showui` and cropping the result, which nobody should have to write twice.

Changes:
- `include_ui` param (default `false`, no behavior change): frames are captured through the engine screenshot path (`FScreenshotRequest` with show-UI — the `shot showui` mechanism), so Slate composites the full backbuffer including UMG before the file is written.
- The write is end-of-frame async, so the per-frame uniformity validation doesn't apply to these frames; they count as requested and the clip directory holds the truth. Documented in the schema.
- Caveat, also documented: with PIE docked in the editor the composited backbuffer is the whole editor window; new-window PIE gives a clean game+UI clip.

Verified on 0.22.0: a 4s `include_ui` clip of a UMG-heavy map produced 7 frames with the full widget stack visible mid-animation — the exact frames the scene-only path renders empty.


tumourlove:
Shipped in v0.23.0.

Reimplemented at HEAD rather than merged, per the usual single-author history. `include_ui` defaults to `false`, frames go through the engine screenshot path when set, and both caveats are documented in the schema: the writes are end-of-frame async so per-frame uniformity is not verified for them (`capture_validity` now says so explicitly, rather than implying a check it did not run), and docked PIE composites the whole editor window.

"Nobody should have to write that twice" was the convincing part of the report. A scene-only path silently returning an empty frame is not something anyone should have to discover by eye.

PR #136, opened 2026-08-11
Add missing includes for non-unity / no-PCH builds

itismyfield:
## What this changes

153 `#include` lines across 106 files. No deletions, no code changes, no behavioural change.

## Why

Unity builds hide missing includes: several `.cpp` files are concatenated into one translation unit, so a file silently gets the declarations it needs from whichever neighbour happened to include them. A shared PCH covers most of the rest.

Building with unity disabled **and** PCHs off surfaces them — 106 files fail to compile because they name types they never include. `-DisableUnity` on its own does not catch these, since the PCH still supplies the headers; that is how a unity-off build gate can stay green while the files are still not self-contained.

This is the include-what-you-use convention the engine's own modules follow, so each fix is mechanical: add the header that declares a type the file already uses.

## Reproducing

In a module's `Build.cs`:

```csharp
PCHUsage = ModuleRules.PCHUsageMode.NoPCHs;
bUseUnity = false;
```

Most frequent additions:

| count | header |
|---|---|
| 21 | `AssetRegistry/AssetData.h` |
| 19 | `Policies/CondensedJsonPrintPolicy.h` |
| 7 | `Dom/JsonObject.h` |
| 6 | `Modules/ModuleManager.h` |
| 5 | `UObject/TextProperty.h`, `UObject/Package.h`, `Serialization/JsonWriter.h`, `Serialization/JsonSerializer.h`, `Math/RandomStream.h` |

By module: MonolithMesh 24, MonolithIndex 21, MonolithCore 12, MonolithAI 9, MonolithLogicDriver 7, MonolithUI 6, MonolithGAS 5, MonolithAudio 5, remainder 1–3 each.

## Two placements worth a look

Both files already had the header, but in a preprocessor branch that does not cover the use site — so these went inside a branch rather than at the top of the file:

- **`MonolithMeshHandlePool.cpp`** — `ListHandles()` in the `!WITH_GEOMETRYSCRIPT` stub branch returns `MakeShared<FJsonObject>()`, while the file's only `Dom/JsonObject.h` sits in the `#else`. Added inside the stub branch, to avoid a redundant include on the GeometryScript path.
- **`MonolithLogicDriverInternal.cpp`** — `AssetRegistry/AssetData.h` added inside `#if WITH_LOGICDRIVER`, alongside the existing include group, since every use site is in that block.

## Verification

The bulk of the set was established against v0.21.2 by compiling each module with unity off and PCHs off on UE 5.7, iterating until clean, then rebased onto `master`. The rebase touched one file — `MonolithBlueprintActions.cpp`, where a later commit added an include at the same position.

Two of the 106 are new files on `master` rather than carried forward, and they are direct evidence for the PCH point above: `MonolithIndexRecoveryTests.cpp` and `MonolithProjectSearchTests.cpp` both call `FPlatformFileManager::Get().GetPlatformFile().DeleteFile(...)` while `HAL/PlatformFileManager.h` only forward-declares `IPlatformFile`, so a no-PCH build fails them with

```
error C2027: use of undefined type 'IPlatformFile'
```

Both now include `GenericPlatform/GenericPlatformFile.h`, matching what `MonolithIndexDatabase.cpp` and `MonolithSourceDatabase.cpp` need for the same call. These two were found by compiling `master`'s sources; the remaining 104 were not recompiled on `master`, so read those as "verified on 0.21.2 / UE 5.7, carried forward mechanically". No 5.8 build was run.

If you would rather have a full no-PCH pass on `master` before merging, say so and I will redo it.

### DIFF
--- a/Source/MonolithAI/Private/MonolithAIBehaviorTreeActions.cpp
+++ b/Source/MonolithAI/Private/MonolithAIBehaviorTreeActions.cpp
@@ -1,4 +1,6 @@
 #include "MonolithAIBehaviorTreeActions.h"
+#include "UObject/TextProperty.h"
+#include "UObject/UnrealType.h"
 #include "MonolithParamSchema.h"
 #include "MonolithAssetUtils.h"
 #include "MonolithJsonUtils.h"
--- a/Source/MonolithAI/Private/MonolithAIControllerActions.cpp
+++ b/Source/MonolithAI/Private/MonolithAIControllerActions.cpp
@@ -1,4 +1,5 @@
 #include "MonolithAIControllerActions.h"
+#include "Editor.h"
 #include "MonolithParamSchema.h"
 #include "MonolithAssetUtils.h"
 
--- a/Source/MonolithAI/Private/MonolithAIIndexer.cpp
+++ b/Source/MonolithAI/Private/MonolithAIIndexer.cpp
@@ -1,4 +1,6 @@
 #include "MonolithAIIndexer.h"
+#include "AssetRegistry/AssetData.h"
+#include "Policies/CondensedJsonPrintPolicy.h"
 #include "MonolithSettings.h"
 #include "AssetRegistry/AssetRegistryModule.h"
 #include "AssetRegistry/IAssetRegistry.h"
--- a/Source/MonolithAI/Private/MonolithAIModule.cpp
+++ b/Source/MonolithAI/Private/MonolithAIModule.cpp
@@ -1,4 +1,5 @@
 #include "MonolithAIModule.h"
+#include "Modules/ModuleManager.h"
 #include "MonolithToolRegistry.h"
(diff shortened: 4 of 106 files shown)


tumourlove:
Shipped in v0.23.0.

Landed all 153 includes. I land contributor patches as my own commits rather than merging the branch -- I keep the shipped history single-author for release integrity and credit contributors in the release notes. No reflection on the patch, which was clean and applied without a conflict.

Your point about the gate was the valuable part, and it turned out to be worse than stated. The release runs a `-DisableUnity` pass specifically to catch missing includes, and it has never caught one, because the shared PCH still supplies the headers -- exactly as you said. That is now written down in `MISSING_FEATURES.md` as a known hole rather than a gate anyone should trust.

This also subsumes #126, which fixed the same two `MonolithAudioRuntime` files.

